PR #5, opened 2019-10-31
CS 3342 Grammars And Syntax Assignment

Voltflare:
Nathan Hites
47786509


README.md
```diff
@@ -42,7 +42,8 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S --> rSr | gSg | bSb
+S --> r | g | b
```
pragdave, line 46: Perfect

README.md
```diff
@@ -53,7 +54,9 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n/2) --> O(n)
+
+To validate a correct path generated by the above grammar, we only need to check/store the first half of the elements, as we know that due to the grammar, it will be symmetric and odd. To validate an incorrect path, we will need to store n/2 (besides the middle, since it's always symmetric with itself) elements and compare them with the other half of the path to validate it.
```
pragdave, line 60: Again, an exemplary answer.

README.md
```diff
@@ -82,7 +85,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+sentence := "the" <adjective> <noun> <verb> <adverb>
+    <adjective> := "lazy" <adjective> | "smelly" <adjective> | "lazy" | "smelly" | (empty)
+        <noun> := "dog" | "cat"
+            <verb> := "ate" | "ran"
+                <adverb> := "slowly" | "noisily" | (empty)
```
pragdave, line 92: In line 89, I think you only need

    <adjective> := "lazy" <adjective> | "smelly" <adjective> | (empty)

# -1

README.md
```diff
@@ -91,7 +98,11 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+sentence := "the" {<adjective>} <noun> <verb> [<adverb>]
+    <adjective> := "lazy" <adjective> | "smelly" <adjective> | "lazy" | "smelly"
```
pragdave, line 102: # -2

You've already handled the zero-or-more part when you wrote {adjective}. So now you just need

README.md
```diff
@@ -109,7 +120,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+See "2_3.jpg"
```
pragdave, line 123: I don't think state S2 is needed: you can transition from S1 back to S1 when you see "lazy" or "smelly"

# -1

README.md
```diff
@@ -125,7 +136,25 @@ Current state | Next word | Next state
...
+    S2          dog             S3
+    S2          cat             S3
+    S3          ate             S4
+    S3          ran             S4
+    S4          slowly          S5
+    S4          noisily         S5
+    S4          EOI             END
+    S5          EOI             END
+    
+
```
pragdave, line 157: # 6

README.md
```diff
@@ -145,7 +174,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+See "main.cpp" for source code. There is also a .pro, .o, and Makefile attached.
```
pragdave, line 177: First, a couple of points that I'm not deducting for.

You included a Makefile that had hard coded paths to stuff on your system. I can't use that.  The .o file also confused things: object files aren't portable.

In the end, I built using 

~~~
 env CPPFLAGS=-std=gnu++11 make main
c++  -std=gnu++11   main.cpp   -o main
main.cpp:98:1: warning: control may reach end of non-void
      function [-Wreturn-type]
}
~~~

When I run it, I get

~~~
14∙41∙32≻ ./main
remember, true = 1 and false = 0
The proper string should be 1010000
Result of Test 1
0

Result of Test 2
0

Result of Test 3
0

Result of Test 4
0

Result of Test 5
0

Result of Test 6
0
~~~

This doesn't seem correct: the comment says I was expecting 101000, so it seems like the code isn't working.

But these aren't really unit tests. A unit test doesn't output stuff that has to be read manually. A unit test is an automated test: it validates expected results against actual, and then reports a single SUCCESS or FAILURE.

The code itself does not really implement a state machine. Instead, it's a linear, manual parse, which only uses the state as a kind of flag to control a couple of loops.

I've attached below a snippet of my C++ implementation

# 4

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

README.md
```diff
@@ -164,4 +193,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Type 1: This particular language exhibits a lack of bunding on the right side of the expression (a feature that begets Type 1), as well as knowledge about the current state being relevant in determining what state to next go to (another context-sensitive feature).
```
pragdave, line 196: # 0

The fact that it can be represented using a pure state transition grammar means it is a type 3 grammar.

The context sensitivity of a type 1 grammar refers to the left-hand side: the nonterminal has to be interpreted in context.

PR #7, opened 2019-10-31
Patrick DeVries

PatrickDeVries:
Patrick DeVries
47381598


README.md
```diff
@@ -40,9 +40,16 @@ Write a Chomsky type 2 grammar that describes any valid path containing at
...
+
 ## A1.1
 
-«replace this with your answer»
+S -> A 
+color -> R | G | B
+A -> color | R A R | G A G | B A B
+R -> 'r'
+G -> 'g'
+B -> 'b'
```
pragdave, line 52: # 5

I'm not sure why you introduced the extra nonterminals, but it doesn't affect the grammar.

README.md
```diff
@@ -53,7 +60,9 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+I believe it would be O(n) because the parser could have one variable pointing to the head
```
pragdave, line 63: # 1
(for the O(n))

The question was about memory use, not CPU.

README.md
```diff
@@ -82,7 +91,26 @@ Write the BNF (not EBNF) description for this language.
...
+            "cat"
+
+<verb> ::=  "ate" |
+            "ran"
+
+<adv> ::= ""       | 
+          <adverb>
+  
+<adverb> ::=  "slowly" |
+              "noisily"
```
pragdave, line 113: # 7

README.md
```diff
@@ -91,7 +119,19 @@ Write this grammar using EBNF with common extensions
...
+                "smelly"
+
+<noun> ::=  "dog" | 
+            "cat"
+
+<verb> ::=  "ate" |
+            "ran"
+  
+<adverb> ::=  "slowly" |
+              "noisily"
```
pragdave, line 134: # 5

README.md
```diff
@@ -109,8 +149,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+FSM.pdf
```
pragdave, line 152: # 6

README.md
```diff
@@ -125,7 +164,20 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ate    |     S3
+    S2        |    ran    |     S3
+    S3        |   slowly  |     S4
+    S3        |  noisily  |     S4
+    S3        |   'EOI'   |     END
+    S4        |   'EOI'   |     END
+
```
pragdave, line 180: # 6

README.md
```diff
@@ -145,17 +197,16 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
-
+run (with python installed):
+python SentenceChecker.py
```
pragdave, line 201: Thank you for making this table driven. 

There's a small bug, and it's due to the way you chose to drive the FSM. Normally, you view an FSM as a box into which you feed events. That means that the outer loop in this case would be an iterator over the words, and each iteration would trigger a state transition. Instead, your code uses the FSM to consume the input. And that's where the bug crept in: if there are events _after_ the EOI, they will be ignored, as the loop stopps when the state machine is done. 

In this particular case that's not really an issue, because EOI should be the last thing. But, in the general case, it would definitely cause problems. Always have the events drive the FSM.

# -1

README.md
```diff
@@ -145,17 +197,16 @@ code, include a script or makefile that will do the job.
...
 
 ## Q2.6 (3 points)
 
 How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
-
+An infinite amount since there can be as many adjectives as one wants.
```
pragdave, line 209: # 3

README.md
```diff
@@ -164,4 +215,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest Chomsky grammar level for this language would be a type-2 (context-free) laguage. This is because every nonterminal can be expressed as a list of terminals and nonterminals without needing context for the LHS nonterminal.
```
pragdave, line 218: # 0

But it can be represented as an FSM with no stack., which makes it type 3.

PR #8, opened 2019-10-31
Chomsky Grammar Homework - Ethan Potthoff

ethanpotthoff:


README.md
```diff
@@ -42,7 +42,15 @@ represent the tiles.
...
-«replace this with your answer»
+S -> C
+
+S -> C S C
+
+C -> "r"
+
+C -> "g"
+
+C -> "b"
```
pragdave, line 53: You're on the right track, but the rule "S -> C S C" doesn't do what you want, because the two Cs could match different colors.

# 2

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

README.md
```diff
@@ -53,7 +61,8 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+The memory requirement for the parser would be O(1) because at any given time, the parser is comparing just the first and last elements of the path, not the whole path itself (n).
```
pragdave, line 64: # 0

Even if that was what the parser was doing, it would still be O(n), because it has to hold the whole string. But a type 2 grammar like this is parsed using a pushdown stack and a table of transitions, and the stack is where it remembers the tiles.  That stack would be O(n).

README.md
```diff
@@ -82,7 +91,17 @@ Write the BNF (not EBNF) description for this language.
...
+
+<description> ::= <adjective> <description> | <adjective>
+
+<adjective> ::= "lazy" | "smelly"
+
+<noun> ::= "dog" | "cat"
+
+<verb> ::= "ate" | "ran"
+
+<adverb> ::= "slowly" | "noisily"
```
pragdave, line 104: # 7

In general, you'd probably want to factor the optional things out of the <sentence> rule:

    <sentence> ::= "the" <adjectives> <noun> <verb> <optional-adverb>
    <adjectives> ::= <adjectives> <adjective> | «empty»
    <optional_adverb> ::= "slowly" | "noisily" | «empty»

README.md
```diff
@@ -91,7 +110,15 @@ Write this grammar using EBNF with common extensions
...
-«replace this with your answer»
+<sentence> ::= "the" <adjective>* <noun> <verb> [<adverb>]
+
+<adjective> ::= "lazy" | "smelly"
+
+<noun> ::= "dog" | "cat"
+
+<verb> ::= "ate" | "ran"
+
+<adverb> ::= "slowly" | "noisily"
```
pragdave, line 121: # 5

README.md
```diff
@@ -109,7 +136,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+See FSM.pdf in GitHub folder
```
pragdave, line 139: # 6

Nice looking diagram :)

README.md
```diff
@@ -125,7 +152,19 @@ Current state | Next word | Next state
...
+    S1        |     lazy        |       S1
+    S1        |     smelly    |       S1
+    S1        |     dog        |       S2
+    S1        |     cat         |       S2
+    S2        |     ate         |       S3
+    S2        |     ran         |       S3
+    S3        |     slowly    |       S4
+    S3        |     noisily    |       S4
+    S3        |     EOI        |       END
+    S4        |     EOI        |       END
```
pragdave, line 167: # 6

README.md
```diff
@@ -145,7 +184,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+Solution is in "main.cpp" in file. Compile and run "main.cpp" to run unit tests.
```
pragdave, line 187: # 9

It passes its tests, but there isn't a state machine in there. It's all linear code with "if"s.

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
        { 5, {{ "EOI", END }}}
};


        if (t.count(nextWord))
                return t[nextWord];
        else
                return -1;
}
~~

README.md
```diff
@@ -154,7 +193,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+Because there is no limit on the number of adjectives a sentence can have, there are an infinite number of sentences in this language.
```
pragdave, line 196: # 3

Indeed there are

README.md
```diff
@@ -164,4 +203,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest Chomsky grammar level for this language is level 3 because no backtracking or stacks are needed to interpret this language (type 2) and the language can be easily represented using a finite state machine.
```
pragdave, line 206: # 3

Perfect answer

PR #10, opened 2019-10-31
Programming Languages Grammar HW Xiyuan Cui

BigKing233:


README.md
```diff
@@ -42,8 +42,13 @@ represent the tiles.
...
 
-«replace this with your answer»
-
+S -> <<empty>>
+S -> R S R
+S -> G S G
+S -> B S B
+R -> r
+G -> g
+B -> b
```
pragdave, line 51: # 3

Close, but your grammar generates even length strings, not odd.

README.md
```diff
@@ -53,7 +58,10 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O((n+1)/2) => O(n)
+In the general, the parser needs to store half of the path configuration. 
+Being symmetric, we only need to parse half of the tiles to parse the whole input.
+After the parser "read" the first half plus the middle one, the whole sentence is done.
```
pragdave, line 64: # 3

(although it doesn't really work that way: there's no way to know you're 1/2 way through...)

README.md
```diff
@@ -82,7 +90,17 @@ Write the BNF (not EBNF) description for this language.
...
+               "The" <Nouns> <Verbs> <Adverbs> |
+               "The" <Adjectives> <Nouns> <Verbs> |
+               "The" <Adjectives> <Nouns> <Verbs> <Adverbs> |
+               "The" <Adjectives> <Adjectives> <Nouns> <Verbs> |
+               "The" <Adjectives> <Adjectives> <Nouns> <Verbs> <Adverbs>
+<Adjectives> ::= " lazy" | " smelly"
+<Nouns> ::= " dog" | " cat"
+<Verbs> ::= " ate" | " ran"
+<Adverbs> ::= " slowly" | " noisily"
+
```
pragdave, line 103: # 7

Two comments.

- it's traditional to use lower case names between the <>s
- rather than list every permutation of sentence, you could use BNFs ability to express the idea of optional items (using an empty production)

README.md
```diff
@@ -91,7 +109,11 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+<sentence> ::= "The" <Adjectives>* <Nouns> <Verbs> <Adverbs>?
+<Adjectives> ::= " lazy" | " smelly"
+<Nouns> ::= " dog" | " cat"
+<Verbs> ::= " ate" | " ran"
+<Adverbs> ::= " slowly" | " noisily"
```
pragdave, line 116: # 5

README.md
```diff
@@ -109,7 +131,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSM_diagram.png
```
pragdave, line 134: # 6

beautiful

README.md
```diff
@@ -125,7 +147,19 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |     S1
+    S1        |    lazy   |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ate    |     S3
+    S2        |    ran    |     S3
+    S3        |   slowly  |     S4    
+    S3        |  noisily  |     S4
+    S3        |    EOI    |    END
+    S4        |    EOI    |    END
```
pragdave, line 162: # 6

README.md
```diff
@@ -145,7 +179,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+SentenceTesting.cpp
```
pragdave, line 182: # 3

You've clearly tried to express the logic of the FSM in code. I was hoping for the actual transitions to be a data structure, though. That's why I asked you to create a table in the previous question. An example of that is at the end of this comment.

The code you supplied didn't compile. I had to add 

~~~ cpp
void tester(string input[],int size);
int eachWord(string word,int condition);
~~~

to fix forward references. Then it didn't run correctly. The main loop does not update the condition.

~~~ cpp
    while(counter<size){
        cout << "condition: " << condition << "  word: " << input[counter] << "\n";
        eachWord(input[counter],condition);
        counter ++;
    }
~~~

Finally, the parse of [ "dog", "EOI" ] falls off the bottom of the switch statement in `eachWord`. The compiler actually warns you about this bug.





Here's an example of using a table for the FSM.

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

README.md
```diff
@@ -154,7 +188,13 @@ How many valid sentences are there in this language?
...
 ## A2.6
 
-«replace this with your answer»
+If all conditions can appear once, 
+which means each adjective can only be used once in each sentence, 
+there will be 60 possible sentences in this language.
+
+But, if we can use adjective more than once, 
+which is allowed by the table and the diagram but banned in the code, 
+then there will be infinitly many sentences in this language.
```
pragdave, line 197: # 3

Correct (although I don't see where it is banned in the code)

README.md
```diff
@@ -164,4 +204,6 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Type-3 grammar will be the simplest Chomsky Grammar level.
+To make the sentence sensable and understandable, we need the terminal signal after each word, which as unterminal; 
+beside, we also need to limite that for each unterminal value, it won't repeat infinitly.
```
pragdave, line 209: # 2

It is type 3, but I don't understand your answer. The simple answer is just "it can be implemented using an FSM with no stack"

PR #12, opened 2019-11-01
CS3342 Grammar Assignment Submission

MorganV:
Morgan VandenBerg
SMU ID: 47760132

Note: my Parser.java code uses the built-in assert keyword, which will only perform tests if the class is run with the -ea flag enabled.


README.md
```diff
@@ -42,8 +42,11 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> A
+S -> S A S
+A -> 'r'
+A -> 'g'
+A -> 'b'
```
pragdave, line 49: # 0

Let's try this grammar. "S -> A" looks good: it generates a single tile.

Now let's try "S -> S A S". Expanding the right hand side, we pick an S, pick an A, then pick an S. Just to help visualize this, I wrote a Ruby program that does this expansion. It randomly chooses a color for A, and it maps S to a single color 60% of the time, and to another round of SAS the rest.

Here's the code:

~~~ ruby
def s()
  if rand() < 0.6
    a()
  else
    [ s(), a(), s() ].join
  end
end

def a()
  ['r', 'g', 'b'].sample
end

10.times do
  puts s
end
~~~

and here's the output:

~~~
17∙47∙50≻ ruby me.rb
rrrgggbrrrgbbbg
rbgbr
r
g
brrggbbbggb
rrbgg
rrrgbgbgggggbrggbgr
g
rrrgbggrg
b
~~~

README.md
```diff
@@ -53,8 +56,12 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+If we calculate space complexity to include the memory that holds the path itself: 
+O(n). It takes n space to hold the path itself, and the path can be validated with a simple loop that does not require external memory.
+Therefore, the total memory complexity O(n). 
 
+If memory complexity is calculated solely for the algorithm which validates the path:
+O(1) because the path can be validated with the same amount of temporary memory no matter the length. The only necessary operation is checking the equality of two array values; space required does not change with size.
```
pragdave, line 64: # 2

Not quite. The first part is correct, but that's not how a parser for this kind of grammar would work. However, you're right that it would storage proportional to the length of the list.

But the second part is cheating. Where are the two copies stored?....

README.md
```diff
@@ -82,17 +89,43 @@ Write the BNF (not EBNF) description for this language.
...
+	     | <adjective> <adjs>
+
+<noun>      ::= "dog"
+	     | "cat"
 
+<adverb>    ::= "slowly"
+	     | "noisily"
+
+<verb>	    ::= "ate"
+	     | "ran"
```
pragdave, line 108: # 7

README.md
```diff
@@ -82,17 +89,43 @@ Write the BNF (not EBNF) description for this language.
...
+             |  "smelly"
+
+<noun>      ::= "dog"
+             |  "cat"
 
+<adverb>    ::= "slowly"
+             |  "noisily"
+
+<verb>      ::= "ate"
+             |  "ran"
```
pragdave, line 128: # 5

I've not seen anyone use the asterisk as a prefix before: it would normally be <adjective>*. But it's a pretty informal notation...

README.md
```diff
@@ -109,8 +142,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+FSM.png
```
pragdave, line 145: # 6

bootiful

README.md
```diff
@@ -125,8 +157,20 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |     S1
+    S1        |   dog     |     S2
+    S1        |   cat     |     S2
+    S2        |   ate     |     S3
+    S2        |   ran     |     S3
+    S3        |   slowly  |     S4
+    S3        |   noisily |     S4
+    S3        |   EOI 
+    S4        |   EOI
+
```
pragdave, line 173: # 6

although you should really show the end state explicity

README.md
```diff
@@ -145,23 +189,21 @@ code, include a script or makefile that will do the job.
...
+Parser.java
 
 ## Q2.6 (3 points)
 
 How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+Because the number of adjectives is not limited and there is no rule against duplicate adjectives, there are infinitely many sentences here.
```
pragdave, line 200: # 3

Indeed there are

README.md
```diff
@@ -145,23 +189,21 @@ code, include a script or makefile that will do the job.
...
 ## Q2.7 (1 point for the level, 2 for the sentence)
 
-Which is the simplest Chomsky grammar level for this language? In one sentence,
-explain why.
+Which is the simplest Chomsky grammar level for this language? In one sentence, explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Chomsky-III. A finite state machine can be used to express the grammar, and it does not require backtracking via a stack.
```
pragdave, line 209: Perfect

# 3

PR #13, opened 2019-11-01
HaydenCenterGrammarHW

HaydenCenter:


README.md
```diff
@@ -42,8 +42,11 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> C
+S -> C S C
+C -> r
+C -> g
+C -> b
```
pragdave, line 50: # 0

The rule "S -> C S C" means "expand S to: choose a C, choose an S, and choose another C". So there's no guarantee that S will start and end the same.

       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b

README.md
```diff
@@ -53,7 +56,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(1), because the amount of memory used at any given operation is simply checking if 2 points are the same, so scaling the sentence does not increase the amount of things to check for each operation, therefore memory is constant.
```
pragdave, line 59: # 0

Which two points? Because it you're assuming that the parser has the whole string in memory, then that's automatically O(n).

README.md
```diff
@@ -82,7 +85,23 @@ Write the BNF (not EBNF) description for this language.
...
+               |  "smelly"
+
+<noun>        ::= "dog"
+               |  "cat"
+
+<verb>        ::= "ate"
+               |  "ran"
+
+<adverb>      ::= "slowly"
+               |  "noisily"
```
pragdave, line 104: # 7

README.md
```diff
@@ -91,7 +110,19 @@ Write this grammar using EBNF with common extensions
...
```
pragdave, line 125: #  5

README.md
```diff
@@ -109,8 +140,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+A2.3.png
```
pragdave, line 143: # 6

README.md
```diff
@@ -125,7 +155,19 @@ Current state | Next word | Next state
...
+    S1        | lazy          |   S1
+    S1        | smelly        |   S1
+    S1        | dog           |   S2
+    S1        | cat           |   S2
+    S2        | ate           |   S3
+    S2        | ran           |   S3
+    S3        | slowly        |   S4
+    S3        | noisily       |   S4
+    S3        | EOI           |   END
+    S4        | EOI           |   END
```
pragdave, line 170: # 6

README.md
```diff
@@ -145,7 +187,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+main.cpp
```
pragdave, line 190: # 10

Nice code, but I was expecting to see the state machine expressed as a table (hence the previous question and the suggestion to "Translate this table into a programming language of your choice" :)

Something like:

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

README.md
```diff
@@ -154,7 +196,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+Since there can be arbitrarily many adjectives, there are an infinite amount of sentences in this language
```
pragdave, line 199: # 3

README.md
```diff
@@ -164,4 +206,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest chompsky grammar type is 3. The language is able to be represented by a finite state machine, and does not require a stack for backtracking
```
pragdave, line 209: # 3

Perfect

PR #14, opened 2019-11-01
Submission for Grammars Assignment

CPetersSMU:
Connor Peters, 47335370


README.md
```diff
@@ -41,8 +41,12 @@ least one paver. Use `S` as the start state, and `r`, `g`, `b` as terminals that
...
 
 ## A1.1
-
-«replace this with your answer»
+S -> r
+S -> g
+S -> b
+S -> r S r
+S -> g S r
+S -> b S r
```
pragdave, line 49: # 5

Yes!

README.md
```diff
@@ -52,8 +56,8 @@ can validate a particular path configuration, where `n` is the number of tiles?
 In one sentence, explain why.
 
 ## A1.2
-
-«replace this with your answer»
+O(n)
+The parser will load the entire path into memory, then check if element i is equal to element n-i, stopping after i = n-i or i = n-i+1, or if element i != n-i thus the memory requirement will be n.
```
pragdave, line 60: # 3

Yes, but... a general purpose parser would not know that it could do that, so it would use a pushdown stack and a state machine. But the answer would be the same: the list of tiles would just end up on the stack.

README.md
```diff
@@ -81,15 +85,22 @@ The following are examples of valid sentences:
...
 
 ## A2.1
-
-«replace this with your answer»
-
+<sentance>::="the"<adjective><noun><verb>|"the"<adjective><noun><verb><adverb>
+<adjective>::= ""|"lazy"|"smelly"|"lazy"<adjective>|"smelly"<adjective>
+<noun>::="dog"|"cat"
+<verb>::="ate"|"ran"
+<adverb>::="slowly"|"noisily"
```
pragdave, line 92: # 7

although line 89 would be better expressed as

~~~
<adjectives>::= "" | <adjectives> <adjective>
<adjective>::=  "lazy" | "smelly"
~~~

(ps. whitespace is your friend...)

README.md
```diff
@@ -81,15 +85,22 @@ The following are examples of valid sentences:
...
 ## Q2.2 (5 points)
 
 Write this grammar using EBNF with common extensions
 
 ## A2.2
+<sentance>::="the"*<adjective><noun><verb>?<adverb>
+<adjective>::="lazy"|"smelly"
+<noun>::="dog"|"cat"
+<verb>::="ate"|"ran"
+<adverb>::="slowly"|"noisily"
```
pragdave, line 103: # 5

This really confused me for a second, because the convention is to put * and ? qualifiers _after_ the term they apply to.

README.md
```diff
@@ -109,7 +120,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSM.png
```
pragdave, line 123: # 6

Very pretty. What tool was that?

README.md
```diff
@@ -125,7 +136,19 @@ Current state | Next word | Next state
...
+    S1        |    lazy   |     S1
+    S1        |   smelly  |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ate    |     S3
+    S2        |    ran    |     S3
+    S3        |   slowly  |     S4
+    S3        |   noisily |     S4
+    S3        |   EOI     |    END
+    S4        |   EOI     |    END
```
pragdave, line 151: # 6

README.md
```diff
@@ -145,7 +168,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+Command to run my code: ./answer.exe
```
pragdave, line 171: Ummm... what's this `.exe` of which you speak? :)

Nice code. I was, however, hoping to see the state transitions expressed in a table (hence the previous question). Something like:

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

# 10

README.md
```diff
@@ -154,7 +177,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+There is an infinite number of valid sentances in this language, as there is no limit on the number of adjectives.
```
pragdave, line 180: # 3

README.md
```diff
@@ -164,4 +187,5 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+This language is a Chomsky Type-3 language.
+Type-3 because the language can be represented with a finite state machine without a stack.
```
pragdave, line 191: # 3

perfect answer

PR #16, opened 2019-11-01
Kirby Cravens Pull Request

kirbycravens:
Kirby Cravens
47320134


README.md
```diff
@@ -42,7 +42,15 @@ represent the tiles.
...
-«replace this with your answer»
+S -> A B A  
+A -> < empty >  
+A -> "r"  
+A -> "g"  
+A -> "b"  
+B -> S  
+B -> "r"  
+B -> "g"  
+B -> "b"
```
pragdave, line 53: # 0

Sorry, but this is way off... I was looking for something like

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

README.md
```diff
@@ -53,7 +61,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n) because the graph of this problem will grow linearly as the length of the path increases.
```
pragdave, line 64: # 1

I don't understand what "the graph...will increase means".

Basically, it it O(n) because you have to keep at least half the input on the stack while processing.

README.md
```diff
@@ -82,7 +90,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+< adjective >   ::=     ""|< a >|< adjective > < a >  
+< a >           ::=     "lazy"|"smelly"  
+< noun >        ::=     "dog"|"cat"  
+< verb >        ::=     "ate"|"ran"  
+< adverb >      ::=     ""|"slowly"|"noisily"
```
pragdave, line 97: # 4

There doesn't seem to be a sentence here.

Also, 

    < adjective >   ::=     ""|< a >|< adjective > < a >  

is actually ambiguous, because <adjective> could be "", so the second and third clauses on the RHS could both match. Better would be
  
    < adjective >   ::=     "" | < adjective > < a >

README.md
```diff
@@ -91,7 +103,10 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+< adjective >     ::=     [{"lazy"|"smelly"}]  
+< noun >          ::=     "dog"|"cat"  
+< verb >          ::=     "ate"|"ran"  
+< adverb >        ::=     ["slowly"|"noisily"]
```
pragdave, line 109: # 3

Again no sentence. And what is `[{"lazy"|"smelly"}]`? {} means zero or more, and [] means optional.

README.md
```diff
@@ -109,7 +124,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSMDiagram.jpeg
```
pragdave, line 127: # 3

The lines between states in the diagram represent events. In your diagram, what event gets me from S1 to S2? And how do I get from S4 to the END is there is no adverb (because if there's no input after the verb, then the event is EOI).

README.md
```diff
@@ -125,7 +140,20 @@ Current state | Next word | Next state
...
+    S1        |    lazy   |     S1  
+    S1        |   smelly  |     S2  
+    S2        |    dog    |     S2  
+    S2        |    cat    |     S3  
+    S3        |    ate    |     S3  
+    S3        |    ran    |     S4  
+    S4        |           |     S4  
+    S4        |   slowly  |     S4  
+    S4        |  noisily  |     S5  
+    S5        |    EOI    |     END
```
pragdave, line 156: # 5

This table seems to be an accurate transcription of your diagram. But the blank entries in the "next word" column should have tipped you off that something was not right...

README.md
```diff
@@ -145,7 +173,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+main.cpp
```
pragdave, line 176: # 3

This code is a correct solution to the wrong problem (which is a shame, because it's a cool idea). 

The question called for you to feed sentences INTO the state machine and see if they matched. Your code generates random sentences based on the grammar.

Also, your generation doesn't really use the state machine properly. It simply iterates through the states, ignoring the next state. As a result, you can only ever get two adjectives out of a run, and they'll always be in the order "lazy" "sme;lly".

README.md
```diff
@@ -154,7 +182,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+48
```
pragdave, line 185: # 0

actually there are an infinite number

README.md
```diff
@@ -164,4 +192,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+This is a level 2 language because its context-free and can be passed down a finite state machine.
```
pragdave, line 195: # 0

It's a type-3 language because it is regular, and can be parsed using a state machine.

PR #18, opened 2019-11-02
Grammars HW

jaymieruddock:
ID 47428255
email jruddock@smu.edu

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,12 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> T
+S -> T S T
+T -> "r"
+T -> "g" 
+T -> "b"
+
 
 
 ## Q1.2  (1 point for the O() answer, 2 for the sentence)
@@ -53,7 +58,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+The likely memory requirement for this is O(1) because it compares only two tiles at any given time so the memory requirement stays the same regardless of n tiles.
 
 
 # Q2
@@ -82,7 +87,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+ ::= "the" | "the" 
+ ::= |
+ ::= "lazy" | "smelly" 
+ ::= "dog" | "cat"
+ ::= "ran" | "ate"
+ ::= "slowly" | "noisily"
 
 
 ## Q2.2 (5 points)
@@ -91,8 +101,11 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
-
+ ::= "the" 
+ ::= { "lazy" | "smelly" } |
+ ::= "dog" | "cat" 
+ ::= "ran" | "ate"
+ ::= [ "slowly" | "noisily" ]
 
 ## Q2.3 (6 points)
 
@@ -125,7 +138,19 @@ Current state | Next word | Next state
 
 ## A2.4
 
-«replace this with your answer»
+Current state | Next word | Next state
+--------------|-----------|-----------
+    S0        |    the    |     S1
+    S1        |    lazy   |   	S1
+    S1	      |   smelly  |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ran    |     S3
+    S2        |    ate    |     S3
+    S3        |   slowly  |     S4
+    S3        |  noisily  |     S4
+    S3        |    EOI    |    END
+    S4        |    EOI    |    END
 
 
 ## Q2.5 (12 points)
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -42,7 +42,12 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> T
+S -> T S T
+T -> "r"
+T -> "g" 
+T -> "b"
+
```
pragdave, line 50: # 3

Very close, but there's no guarantee that both Ts in "T S T" will expand to the same color. Instead you need something like:

       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b

README.md
```diff
@@ -53,7 +58,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+The likely memory requirement for this is O(1) because it compares only two tiles at any given time so the memory requirement stays the same regardless of n tiles.
```
pragdave, line 61: # 0

But where does it store the tiles it is comparing? Remember, we're talking about a parser here. It's a type 2 grammar, so it will need a transition table and a pushdown stack. The list of tiles is going to end up (at least partially) on that stack, so it's O(n).

README.md
```diff
@@ -82,7 +87,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+ ::= "the" | "the" 
+ ::= |
+ ::= "lazy" | "smelly" 
+ ::= "dog" | "cat"
+ ::= "ran" | "ate"
+ ::= "slowly" | "noisily"
```
pragdave, line 95: # 0

A BNF grammar is rules of the form

    <nonterminal> ::= list of terminals or non terminals

There will be a starting nonterminal (in this case <sentence>) from which the rest of the productions can generate the grammar.

README.md
```diff
@@ -91,8 +101,11 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
-
+ ::= "the" 
+ ::= { "lazy" | "smelly" } |
+ ::= "dog" | "cat" 
+ ::= "ran" | "ate"
+ ::= [ "slowly" | "noisily" ]
```
pragdave, line 108: # 0

README.md
```diff
@@ -109,7 +122,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSM.pdf
```
pragdave, line 125: # 4

I'm curious how you arrived at this FSM based on the BNF. 

There's a problem with it, though. You're in S0, and you receive the word "the". Where do you go: S1 or S2? In your diagram it could be either, which makes your state machine nondeterministic.

What should happen is exactly what you wrote in the answer to the next question.

README.md
```diff
@@ -145,7 +170,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+python TSM.py
```
pragdave, line 173: # 7

You produced a parser, but it isn't based on a state machine, and in fact illustrates exactly why a state machine is a good idea. There are flags (such as `noun`) that carry state in an ad hoc way, and there are assumptions (such as the maximum number of adjectives is 2) that aren't true.

A state table approach would look something like:

~~~ py
import unittest
from functools import reduce
ERROR  = "ERROR"
OK     = "OK"

states = {
  0:  { "the":    1 },
  1:  { "lazy":   2, "smelly":  2, "dog": 3, "cat": 3 },
  2:  { "lazy":   2, "smelly":  2, "dog": 3, "cat": 3 },
  3:  { "ate":    4, "ran":     4 },
  4:  { "slowly": 5, "noisily": 5, "EOI": OK },
  5:  { "EOI":    OK }
}

def getNextState(currentState, nextWord):
    return states.get(currentState,{}).get(nextWord, ERROR)

def stateMachine(sentence):
    return reduce(getNextState, sentence, 0) == OK
~~~

README.md
```diff
@@ -125,7 +138,19 @@ Current state | Next word | Next state
...
+    S1        |    lazy   |   	S1
+    S1	      |   smelly  |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ran    |     S3
+    S2        |    ate    |     S3
+    S3        |   slowly  |     S4
+    S3        |  noisily  |     S4
+    S3        |    EOI    |    END
+    S4        |    EOI    |    END
```
pragdave, line 153: # 6

README.md
```diff
@@ -154,7 +179,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+There are infinite valid sentences because it can be any combination of words as there is no specification on length, only validity.
```
pragdave, line 182: # 3

README.md
```diff
@@ -164,4 +189,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest Chomsky grammar level for this language is level 3 because it can be parsed by a finite state machine because the next symbol is always a terminal symbol.
```
pragdave, line 192: # 3

perfect answer

PR #20, opened 2019-11-02
Eric Bernard Homework

mariomanoptic:


README.md
```diff
@@ -42,8 +42,11 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> C
+S -> SCS
+C -> "r"
+C -> "g"
+C -> "b"
```
pragdave, line 49: # 0

Try running this in your head, using it to generate a path of length 3. You'd need to execute 

    S -> S C S

The first S on the RHS will have to expand to C, which expands to a color. Let's say "r".
The C will expand to a color, say "g". The final S must expand to C, which expands to (say) "g". So we have "rgb", which isn't valid.

Or, we could try again, but have the last S expand itself to SCS. That would lead to "rgrgb".

       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b

README.md
```diff
@@ -53,7 +56,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+the parser would only need O(1) memory because it will always check either one or two values regardless of the number of overall values
```
pragdave, line 59: # 0

And where are those values stored?

Remember that the parser works using a pushdown stack and a transition table. The list of tiles will need to be stored on that stack (at least partially) so the answer is O(n).

README.md
```diff
@@ -82,7 +85,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+::= "the" | "the" 
+::= | 
+::= "lazy" | "smelly" 
+::= "dog" | "cat" 
+::= "ran" | "ate" 
+::= "slowly" | "noisily"
```
pragdave, line 93: # 0

this is not BNF

README.md
```diff
@@ -91,7 +99,10 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+::= "the" ::= { "lazy" | "smelly" } | 
+::= "dog" | "cat" 
+::= "ran" | "ate" 
+::= { "slowly" | "noisily" }
```
pragdave, line 105: # 0

README.md
```diff
@@ -125,7 +136,17 @@ Current state | Next word | Next state
```
pragdave, line 136: # 6

(diagram)

README.md
```diff
@@ -125,7 +136,17 @@ Current state | Next word | Next state
...
+--------------|-----------|-----------
+    S0        |    the    |     S1
+    S1        |   noun    |     s2
+    s1        |  adjective|     s3
+    s3        | adjective |     s3
+    s3        |   noun    |     s2
+    s2        |   verb    |     s4
+    s4        |   EOI     |     s6
+    s4        |   adverb  |     s5
+    s5        |   EOI     |     s6
```
pragdave, line 149: # 6

README.md
```diff
@@ -154,7 +175,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+48
```
pragdave, line 178: # 0

There are an infinite number, as you can do the S3 -> S3 transition for as long as there's an adjective in the input.

README.md
```diff
@@ -164,4 +185,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+level 3 because as we displayed the language can be parsed using a finite state machine thus making it a level 3 chomsky grammar
```
pragdave, line 188: # 3

README.md
```diff
@@ -125,7 +136,17 @@ Current state | Next word | Next state
...
+    s1        |  adjective|     s3
+    s3        | adjective |     s3
+    s3        |   noun    |     s2
+    s2        |   verb    |     s4
+    s4        |   EOI     |     s6
+    s4        |   adverb  |     s5
+    s5        |   EOI     |     s6
 
 
 ## Q2.5 (12 points)
```
pragdave, line 152: # 9

There are two issues here. The question asks you to translate the state table into your language, so I was expecting to see it expressed as a data structure, perhaps something like this:

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

Second, the code in interpreter (which is clearly laid out, btw) has a bug. It exits the while loop early when it gets to the EOI. In this case, that also corresponds to the end of the string, but in the general case that's not guaranteed. This is a common error: code that stops after it finds that it has what it is looking for, rather than checking that all it has is what it is looking for.

PR #26, opened 2020-04-01
all questions attempted or completed

sjkaiser18:
Sarah Kaiser 47349124


README.md
```diff
@@ -41,8 +41,23 @@ least one paver. Use `S` as the start state, and `r`, `g`, `b` as terminals that
...
+
+S -> L
+
+R -> "r"
+
+G -> "g"
+
+B -> "b"
+
+L-> "r"|"g"|"b"
```
pragdave, line 60: Before I looked into this, I had to tidy it up a little. I removed the blank lines, and replaced the R,G,and B NTs with terminals:

~~~
S -> < empty >
S -> r S r
S -> g S g
S -> b S b
S -> L
L->  r|g|b
~~~

I then removed the L NT

~~~
S -> < empty >
S -> r S r
S -> g S g
S -> b S b
S -> r|g|b
~~~

And suddenly all is clear. And you're so, so close. The only problem is that very first line. Let's generate a production using `r S r`, and choose `S -> <empty> `. That gives us a result of `rr`, which is an even length string.

I suspect you did this deliberately, in order to allow both even length and odd length paths which are palidromic. But technically it is not what the question wanted.

# -1

README.md
```diff
@@ -53,8 +68,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+O(n), since esentially it is determining a "palindrome" persay, so the maximum is n/2 because it may be false before then, and since n/2 is still greater than a linear search, it is still O(n).
```
pragdave, line 71: # -1

The O(n) is right, but remember we're talking about memory here, and any memory used to parse this is in a stack, so that should be the focus of the answer.

README.md
```diff
@@ -82,7 +96,19 @@ Write the BNF (not EBNF) description for this language.
...
+
+< noun >::="dog"|"cat"
+
+< adjective >::="smelly"|"lazy"
+
+< predicate >::=< verb >|< verb >< adverb >
+
+< verb >::="ate"|"ran"
+
+< adverb >::="slowly"|"noisily"
```
pragdave, line 111: # -3

Unfortunately this grammar doesn't allow for repeating adjectives or optional adverbs.

README.md
```diff
@@ -109,8 +142,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+![If the image will not appear, view fsm.jpg included in this folder](/fsm.jpg)
```
pragdave, line 145: Good diagram

README.md
```diff
@@ -125,8 +157,19 @@ Current state | Next word | Next state
...
+  2    |      S1        |    dog         |    S2
+  3    |      S1        |    cat         |    S2
+  4    |      S1        |    smelly      |    S1
+  5    |      S1        |    lazy        |    S1  
+  6    |      S2        |    ate         |    S3
+  7    |      S2        |    ran         |    S3
+  8    |      S3        |    slowly      |    S4
+  9    |      S3        |    noisily     |    S4
+  10   |      S3        |      .         |    END
+  11   |      S4        |      .         |    END
```
pragdave, line 172: Nice presentation

README.md
```diff
@@ -145,16 +188,27 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+run main.cpp with sentences.txt as a command line argument:
```
pragdave, line 191: It's a nice piece of code, but a couple of things are holding it back.

1. It doesn't have any tests. Yes, it does take input from a file and spray messages to the console, but that doesn't test the code; at best it lets some hapless human do it. A test knows the expected result from each call, and reports if the actual result is different.

# -2

Second, rather than use the state table as a data structure, you wrote it all in executable code. Normally that's a 6 point deduction, but your code actual is a representation of the states, so it's -3

# -3

README.md
```diff
@@ -145,16 +188,27 @@ code, include a script or makefile that will do the job.
...
 How many valid sentences are there in this language?
 
 ## A2.6
+There are an __infinite__ amount of solutions
+
+mathematical explanation:
+
+The smelly/lazy/blank* dog/cat ate/ran slowly/noisily/blank .
+
+1 * infinity * 2 * 2 * 3 * 1=infinity
```
pragdave, line 209: Must be true 'cos math!

PR #27, opened 2020-04-03
Homework 09

cuom1999:
Name: Phuoc Dinh Le
SMU ID: 47601532


README.md
```diff
@@ -42,7 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+    S -> r | g | b
+    S -> r S r
+    S -> b S b
+    S -> g S g
```

README.md
```diff
@@ -53,7 +56,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n). Because it should check if the configuration is an odd-length palindrome; one way to do it is using one pointer goes from left to right and one from right to left, they should meet at the center so we need n / 2 comparisions, which is O(n).
```
pragdave, line 59: # -2

The answer is O(n), but you've considered runtime speed, not memory use. In fact, the runtime speed for a parser is likely to be between O(n<sup>2</sup>) and O(n<sup>3</sup>).

README.md
```diff
@@ -109,8 +125,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+![](diagram.jpg)
```
pragdave, line 128: Good job: you worked out that adjectives keep the machine in state 1.

README.md
```diff
@@ -145,16 +174,17 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+Command: g++ -std=c++14 sentence.cpp -o sentence && ./sentence
```
pragdave, line 177: Thank you, thank you, for restoring my faith in C++. This is one of the few C++ submissions that didn't kill off a small corner of my soul.

# +3 bonus

You had good tests, both positive and negative, and you used a table to control the state machine.

pragdave:
Very nice job.

PR #28, opened 2020-04-04
Struggled on 2.5

itrose225:


README.md
```diff
@@ -42,7 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> RSR | GSG | BSB | R | G | B
+R -> "r"
+G -> "g"
+B -> "b"
```
pragdave, line 48: This is correct. For me, I find it clear to express it as

S -> r | g | b | rSr | gSg | bSb

The reason is that it R is a nonterminal, then it can potentially take different values if it appears twice in a production. That's not the case here, but it was still a little confusing when I first read it.

README.md
```diff
@@ -53,7 +56,8 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n)
+It has to be able to compare the first and last tiles for symmetry, along with every other set.
```
pragdave, line 60: # -1

It is O(n), but any discussion of memory use should mention the stack.

README.md
```diff
@@ -82,7 +86,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<sentence>   ::= "the" <adjective> <noun> <verb> <adverb>"."
+<adverb>     ::= "slowly" | "noisily" | ""
+<noun>       ::= "dog" | "cat"
+<verb>       ::= "ate" | "ran"
+<adjecctive> ::= "lazy" | "smelly" | ""
```
pragdave, line 93: Close, but this doesn't allow for repeating adjectives

# -2

README.md
```diff
@@ -109,7 +121,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+filename.jpg
```
pragdave, line 124: `filename.jpg`? srsly?

This is strange: your diagram has the correct topology, but it swaps state names for labels.

So there should be a transition from S0 to S1 labelled "The". And so on.

Here's my version:

image

# -3

README.md
```diff
@@ -120,13 +132,27 @@ Current state | Next word | Next state
...
+    S1        |    ""     |     S2
+    S1        |    lazy   |     S1
+    S1        |    smelly |     S1
+    S2        |    cat    |     S3
+    S2        |    dog    |     S3
+    S3        |    ran    |     S4
+    S3        |    ate    |     S4
+    S4        |    slowly |     eoi
+    S4        |    noisily|     eoi
+    S4        |    ""     |     eoi
```
pragdave, line 155: This table doesn't really agree with your diagram. For example, you diagram has two transitions from S0, and this table only has one.

This table also has transitions marked with an empty string. That can't work, because the state machine would never know when to apply it. For example, it's in S1. The next word is "lazy." The state machine _could_ choose it. Or it could simply say "I'll match nothing and go to S2", breaking the parse.

# -3

Here's a version of the table for the diagram above

README.md
```diff
@@ -154,7 +180,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+((2*5)*2)*3 -> 60
```
pragdave, line 183: # -2

In a version with no repeated adjectives that might be right. But the actual language has an infinite number of sentences.

README.md
```diff
@@ -164,4 +190,5 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Type 2. It is context-less as no word is dependent on the word before or after it,
+yet is not quite normal as the adjectives and adverbs can be empty.
```
pragdave, line 194: Because we can parse it with a state machine, it is type 3. The next token _always_ determines what we do next, so it is deterministic.

# -3

README.md
```diff
@@ -120,13 +132,27 @@ Current state | Next word | Next state
...
+    S1        |    smelly |     S1
+    S2        |    cat    |     S3
+    S2        |    dog    |     S3
+    S3        |    ran    |     S4
+    S3        |    ate    |     S4
+    S4        |    slowly |     eoi
+    S4        |    noisily|     eoi
+    S4        |    ""     |     eoi
 
 ## Q2.5 (12 points)
```
pragdave, line 157: Your code was on the correct path. You had a data structure representing the table. You just needed to change it slightly to allow it to be traversed. And if you had the correct state table to base it on, you'd find your code would become a lot easier.

Here's a possible version:

~~~ c++
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
// current    transitions
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}

    int state = 0;
    for (string word: words) {
        if (state < 0)
            return false;
    }
    return state == 5;
}
~~~

# -8

PR #29, opened 2020-04-04
Finished Grammars HW

pattisonb:
finished all the questions


README.md
```diff
@@ -53,7 +53,8 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+0(1)
+You can compare each character in the first half to its mirror in the second half to check equality and this way you would only ever have two tiles in memory at the same time.
```
pragdave, line 57: You could if you had everything available. But that's not a parser. 

# -3

README.md
```diff
@@ -82,7 +83,15 @@ Write the BNF (not EBNF) description for this language.
...
-«replace this with your answer»
+\<sentence> ::= "the" + \<adjective> + \<noun> +  \<verb> + \<adverb>
+
+\<adjective> ::= \<<empty>> | "lazy" | "smelly"
+
+\<noun> ::= "dog" | "cat"
+
+\<verb> ::= "ate" | "ran"
+
+\<adverb> ::= \<<empty>> | "slowly" | "noisily"
```
pragdave, line 94: this is almost correct; it just doesn't allow for multiple adjectives.

# -2

README.md
```diff
@@ -109,7 +126,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSM_Diagram.png
```
pragdave, line 129: The transition from S3 to S4 on the empty string isn't correct: there can't be a transition that isn't triggered by something. Instead S4 and S3 both have to deal with EOI:

image

# -2

README.md
```diff
@@ -125,7 +142,20 @@ Current state | Next word | Next state
...
+    S0        |    the        |     S1
+    S1        |   lazy	      |   	S1
+    S1        |   smelly      |   	S1
+    S1        |   dog         |     S2
+    S1        |   cat         |     S2
+    S2        |   ate         |    	S3 
+    S2        |   ran         |     S3
+    S3        |    noisily    |     S4
+    S3        |    slowly     |     S4
+    S3        |   <<empty>>   |   	S4
```
pragdave, line 156: Same comment here, but no deduction because the table reflects the diagram.

README.md
```diff
@@ -145,7 +175,9 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+The file is called sentences.cpp.
```
pragdave, line 178: I appreciate you adding code to test both positive and negative cases. However, these are actually tests, as what they do is write 1s and 0s to the console. A test will check that the code returns what is expected, and report when it doesn't.

Here's what I mean, with a fairly minimal hack to your code.

~~~ c++
using namespace std;
typedef vector<sentence> sentences;
// ...
int main() {
	sentences valid {
	  {"the", "dog", "ate", "EOI"},
	  {"the", "lazy", "smelly", "dog", "ran", "slowly", "EOI"},
	  {"the", "lazy", "smelly", "lazy", "dog", "ate", "noisily", "EOI"},
	  {"the", "dog", "ate", "slowly", "EOI"},
	  {"the", "dog", "ran", "noisily", "EOI"},
	};
	 sentences invalid {
	  {"dog", "ate", "EOI"},
	  {"the", "lazy", "smelly", "ran", "EOI"},
	  {"the", "lazy", "smelly", "dog", "ate", "slowly", "noisily", "EOI"},
	  {"the", "dog", "EOI"},
	 };

	for (sentence s: valid) {
		for (const auto &e : s) cout << e << " "; cout << endl;
		assert(isSentence(s));
	}
	for (sentence s: invalid) {
		for (const auto &e : s) cout << e << " "; cout << endl;
		assert(!isSentence(s));
	}

	return 0;
};
~~~

# -2

Now on to the parser.

The idea of the exercise was to "Translate this table into a programming language of your choice"; that is to create an actual data structure in C++ and then traverse it during the parse. The next best alternative would be to have written the state machine totally in code, with (maybe) a function per state, or a case statement. But the code here is just a manually written parser without any mention of states. It's actually a good example of why state machines are helpful. This code has if statements nested 7 level deep. Here's a version using a table:

~~~ c++
typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

# -6

README.md
```diff
@@ -164,4 +196,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Type 2 because the langauge can repeat adjectives making it infite therefore it cannot be Type 3.
```
pragdave, line 199: But is can be parsed using a state machine, which makes it type 3.

# -3

PR #31, opened 2020-04-05
finished assignment

amoryweinzierl:


README.md
```diff
@@ -42,8 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> r | g | b
+S -> r S r
+S -> g S g
+S -> b S b
```

README.md
```diff
@@ -53,8 +55,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+O(n). The parser has to check each elements and pair ends together, which is O*n/2), but then, ignoring ignoring the factor of division, it would be O(n)
```
pragdave, line 58: It _is_ O(n), but I think you're doing the analysis for speed, and not memory usage...

# -1

README.md
```diff
@@ -82,7 +83,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<sentence> ::= "the" <adjectives> <noun> <verb> <adverb>
+<adjectives> ::= <adjectives> <adjective> | <adjective>
+<adjective> ::= "lazy" | "smelly" | ""
+<noun> ::= "dog" | "cat"
+<verb> = "ate" | "ran"
+<adverb> = "slowly" | "noisily" | ""
```
pragdave, line 91: No deduction, but my spidey sense is tingling with the lines:

~~~
<adjectives> ::= <adjectives> <adjective> | <adjective>
<adjective> ::= "lazy" | "smelly" | ""
~~~

I wonder if this might loop forever because <adjectives> can repeatedly call <adjective> matching nothing. I think it would be safer to have

~~~
~~~

README.md
```diff
@@ -109,8 +118,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+FSMdiagram.jpg
```
pragdave, line 121: Love the hairy circles :)

Good job noticing that you can loop back to S1 on adjectives.

README.md
```diff
@@ -145,17 +165,16 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
-
+files to test are parser.h, parser.cpp, main.cpp
+run main.cpp and it will input true and false based on tests declared in the main
```
pragdave, line 169: First off, thank you for writing both positive and negative tests. My only comment is that they're not really tests: they just run code out output stuff, leaving it for a poor human to work out whether the code is good. A unit test will run the code and report if the value returned is not the expected value.

I quickly hacked your code:

~~~ c++
#include <iostream>
#include <assert.h>

#include "parser.h"
typedef vector<sentence> sentences;

using namespace std;
int main(){
    sentences valid {
      {"the", "lazy", "cat", "ate", "slowly", "EOI"},
      {"the", "lazy", "dog", "ate", "slowly", "EOI"},
      {"the", "smelly", "cat", "ran", "slowly", "EOI"},
      {"the", "smelly", "dog", "ran", "noisily", "EOI"},
      {"the", "smelly", "cat", "ate", "noisily", "EOI"},
    };

    sentences invalid {
      {"lazy", "cat", "ate", "slowly", "EOI"},
      {"dog", "ate", "slowly", "EOI"},
      {"the", "smelly", "ran", "slowly", "EOI"},
      {"the", "smelly", "dog", "ran", "noisily"},
      {"EOI"},
    };

    parser parse;

    for (sentence s: valid) {
        assert(parse.FSManalyze(s));
    }

    for (sentence s: invalid) {
        assert(!parse.FSManalyze(s));
    }
}
~~~

# -2

On to the parser.

First, I have to say that your code was a joy to read. I've had a lifetime's worth of C++ in the last two days, and this was this chunk of code that didn't make a part of my soul curl up a little. So 3 bonus points from my soul.

# +3

However, you actually didn't do what the assignment wanted: I was hoping you'd take the table from above and transcode it into a C++ data structure (`map<map<int,string>, int>`) and then work from that. At the same time, you did actually use states in your code, so rather than deducting 6 for this, I'm deducting 3

# -3

FYI: here's the code using a table

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

pragdave:
Very nice!

PR #32, opened 2020-04-05
Grammars HW

fidelianawar-zz:
Fidelia Nawar - 47963894


README.md
```diff
@@ -42,7 +42,13 @@ represent the tiles.
...
 ## A1.1
 
-«replace this with your answer»
+S -> r S r
+
+S -> g S g
+
+S -> b S b
+
+S -> r | g | b
```
pragdave, line 51: Perfect

README.md
```diff
@@ -53,7 +59,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+The time complexity would be O(n) because looking for a palindrome is 0(n/2). This is because only parsing half of the path is necessary to determine if it is a palindrome and O(n/2) is equivalent to O(n) because with a large 'n,' the 1/2 component does not matter.
```
pragdave, line 62: Actually, the time complexity is at least O(n<sup>2</sup>), and might go as high as O(n<sup>3</sup>). But I asked for the memory usage, which is indeed O(n).  The parser takes its input one token at a time. It doesn't know where the end is

# -1

README.md
```diff
@@ -82,7 +88,19 @@ Write the BNF (not EBNF) description for this language.
...
+
+\<action> ::= \<verb> | \<verb> " " \<adverb>
+
+\<noun> ::= "dog" | "cat"
+
+\<adjective> ::= "lazy" | "smelly"
+
+\<verb> ::= "ate" | "ran"
+
+\<adverb> ::= "slowly" | "noisily"
```
pragdave, line 103: Cool: I haven't seen anyone else split it up this way.

The only problem I see is that you don't allow repeated adjectives.

# -2

README.md
```diff
@@ -106,10 +132,10 @@ Write this grammar using EBNF with common extensions
 
   (Hint: my answer has seven states including the start and end states)
 
-
+   
 ## A2.3
 
-«replace this with your answer»
+https://github.com/fidelianawar/cs3342_09/blob/master/Q2.3%20FSM.pdf
```
pragdave, line 138: First of all, you have the best handwriting of any diagram so far!

The diagram is _almost_ right. The only problem is that empty string between S1 and S2. You can't have a transition without an event. But all you need to do is combine S1 and S2:

image


# -2

README.md
```diff
@@ -125,7 +151,19 @@ Current state | Next word | Next state
...
+|   S1        |   lazy    |     S1
+|    S1        |   smelly  |     S1
+|    S1        |    cat    |     S2
+|    S1        |    dog    |     S2
+|    S2        |    ate    |     S3
+|    S2        |    ran    |     S3
+|    S3        |  slowly   |     S4
+|    S3        |  noisily  |     S4
+|   S3        |    EOI    |    End
+|   S4        |    EOI    |    End
```
pragdave, line 166: ... which is what you did in this table. strange

README.md
```diff
@@ -145,7 +183,12 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+To run main.cpp, type these commands:
```
pragdave, line 186: First, thank you for providing both positive and negative tests. The thing that stops them getting full marks is that they don't actually do any testing: they just report facts to the console, and a poor human has to match them against some strings in the code. A test would compare the result of each parse against the expected value, and report when they didn't match.

# -2

You `determineValidity` function is a nice piece of C++ (and I don't say that lightly: I've read enough C++ in the last two days to last a lifetime, so it has to be really good for me to appreciate it.)  However, you aren't actually implementing a state machine. In fact, deleting the variable `state` makes no difference. I was hoping to see a pure table-based approach:

~~~ c++
typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

I normally deduct 6 for not using a table, but given how nice your code was to read:

# -3

README.md
```diff
@@ -164,4 +206,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest grammar level is Type 3 because on the left-hand side, we have a single non-terminal (S) and on the right-hand side we have a single terminal (r | g | b) or a single terminal followed by a single non-terminal.
```
pragdave, line 209: Yup. Also because it can be expressed as a FSM

PR #33, opened 2020-04-05
Finished

DylanCaro22:


cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+Write a Chomsky type 2 grammar that describes any valid path containing at
+least one paver. Use `S` as the start state, and `r`, `g`, `b` as terminals that
+represent the tiles.
+
+## A1.1
+
+
+S -> color1 | color2 S color2
+
+color -> r|g|b
```
pragdave, line 48: # -3

Some problems here: 

1. color, color1, and color2 are terminals, so line 48 is not meaningful, and the only valid outputs of the grammar are "color1" or "color2 x n.... color1 ... color2 x n"

2. If we assume you means color1 and color2 to be nonterminals, they aren't defined

3. If they were defined, `color2 S color2` doesn't do what you think it does, because there's no guarantee that two occurrences of the same nonterminal match the same input. (if they didm rules such as `<number> :- <digit>+` would only match numbers with repeating digits.

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+## Q1.2  (1 point for the O() answer, 2 for the sentence)
+
+Using big-O notation, what is the likely memory requirement for a parser that
+can validate a particular path configuration, where `n` is the number of tiles?
+In one sentence, explain why.
+
+## A1.2
+
+O(1) becuase the parser would only need to iterate over the path one time
```
pragdave, line 58: That's an estimate of speed, not memory.

# -3

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+Write the BNF (not EBNF) description for this language.
+
+## A2.1
+
+\<sentence> ::= "the" + \<adj> + \<noun> + \<verb> + \<adverb>
+\<adj> ::= \<<empty>> | "lazy" | "smelly"
+\<adverb> ::=  \<<empty>> | "slowly" | "noisily"
+\<noun> ::=   "dog" | "cat"
+\<verb> ::=   "ate" | "ran"
```
pragdave, line 91: I don't know what the backslashes are for and I'm ignoring them

The plus signs are not BNF (-1) and the grammar doesn't allow for multiple adjectives (-2)

# -3

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+  You can hand draw it and snap a picture, or use a drawing tool. Either way,
+  upload the image, and put the file name in the answer below.
+
+  (Hint: my answer has seven states including the start and end states)
+
+
+## A2.3
+
+Name = A2.3
```
pragdave, line 122: I think the labels on some of the transitions are incorrect: S1 -> S2 should be "noun", and S2 -> S3 should be verb. Then S3 should go to the END state on EOI, and to S4 on seeing an adverb. S4 should also go to the end state on EOI

image


# -3

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+    S1        |   lazy    |     S2
+    S1        |   smelly  |     S2
+    S1        |   "empty" |     S2
+    S2        |   dog     |     S3
+    S2        |   cat     |     S3
+    S3        |   ran     |     S4
+    S3        |   ate     |     S4
+    S4        | slowly    |     S5
+    S4        | noisily    |     S5
+    S5        | EOI       |     END
```
pragdave, line 151: The diagram is fine past S2 (but you're missing the transition `S4 -> EOI -> END`.

In a state machine you have states (the circles) and directed transitions between the circles. The transitions are each labelled with an event. When you are in a particular state and you receive an event, you look for the transition with that even as a label and follow it to the next state.So every transition must be caused by an event, and «empty» is not an event. If it were, how could the FSM choose what to do in S1.

image

# -3

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+Then write some unit tests that exercise your function, making sure that it
+recognizes valid sentences and rejects invalid ones.
+
+The answer should appear in one or more source files in the same directory as
+this file. If I need to do anything more that type a single command to run your
+code, include a script or makefile that will do the job.
+
+## A2.5
+
+ProgLangHomework.cpp
```
pragdave, line 171: So, here's the code I accidentally revealed last week:

~~~ cpp
typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

and here's yours

~~~ cpp
typedef int state;
typedef std::map<state, transitions> stateTable;
const int END = 99;
stateTable states = {
    {0, {{"the", 1}}},
    {1, {{"lazy", 2}, {"smelly", 2}, {"dog", 3}, {"cat", 3}}},
    {2, {{"lazy", 2}, {"smelly", 2}, {"dog", 3}, {"cat", 3}}},
    {3, {{"ate", 4}, {"ran", 4}}},
    {4, {{"slowly", 5}, {"noisily", 5}, {"EOI", END}}},
    {5, {{"EOI", END}}}};
int nextState(int curState, string nextword) {
  transitions t = states[curState];
  if (t.count(nextword))
    return t[nextword];
  else
    return -1;
}
~~~

There's no way you could get from your previous table to this code, which is identical (down to the names) to mine.

# 0

cs3342_09-master/README.md
```diff
@@ -0,0 +1,191 @@
...
+
+
+## Q2.7 (1 point for the level, 2 for the sentence)
+
+Which is the simplest Chomsky grammar level for this language? In one sentence,
+explain why.
+
+## A2.7
+
+The simplest Chomsky language level for this language would be a level 2 becuase it can go infinitely so it can't be type 3
```
pragdave, line 191: It can be parsed by a state machine, so it is actually type 3.

# -3

PR #34, opened 2020-04-05
Grammar Assignment

johnSMUpark:
Name:    John Park
SMU ID: 446 88 167


README.md
```diff
@@ -42,8 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> r | g | b
+S -> rSr
+S -> gSg
+S -> bSb
```
pragdave, line 48: perfect

README.md
```diff
@@ -53,21 +55,17 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(1). To validate a path, you only need memory for 2 pointers traversing from either ends of the path to the path's midpoint regardless of the number of tiles in the path.
```
pragdave, line 58: The parser takes its input one token at a time. It doesn't know where the end is

# -3

README.md
```diff
@@ -109,7 +116,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+Please open FSM.pdf
```
pragdave, line 119: The diagram is fine past S2, but the first three states are kind of screwy.

You can't have two transitions from a particular state with the same label (because how could the state machine choose between them) and every transition must be caused by an event.

# -3

README.md
```diff
@@ -125,7 +132,22 @@ Current state | Next word | Next state
...
+S1            | smelly    | S1
+S1            | smelly    | S2
+S2            | dog       | S3 
+S2            | cat       | S3
+S3            | ate       | S4
+S3            | ran       | S4
+S4            | slowly    | S5
+S4            | noisily   | S5
+S4            | EOI       | END
+S5            | EOI       | END
```
pragdave, line 150: The same comments apply to this table. When you found yourself writing two transitions with the same label from one state, you should have stopped and wondered what that could actually mean. FSMs are deterministic, so they don't make guesses.

No extra deduction: you tried to represent your diagram

README.md
```diff
@@ -140,12 +162,12 @@ Then write some unit tests that exercise your function, making sure that it
...
 
 The answer should appear in one or more source files in the same directory as
-this file. If I need to do anything more that type a single command to run your
+this file. If I need to do anything more than type a single command to run your
 code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+python fsm.py
```
pragdave, line 170: You used a table,  and you included a nice set of tests, both positive and negative. The only problem was that ambiguity you introduced in the table, which lead to you having to write:

~~~ py
                if len(next)==1: state = next[0]
                else:
                    if words[index + 1]=='dog' or words[index + 1]=='cat': 
                        state = next[1]
                    else: 
                        state = next[0]
~~~

which is clearly not clean.

# -2

(I was looking for a table that looked more like

~~~ py
states = {
  0:  { "the":    1 },
  1:  { "lazy":   2, "smelly":  2, "dog": 3, "cat": 3 },
  2:  { "lazy":   2, "smelly":  2, "dog": 3, "cat": 3 },
  3:  { "ate":    4, "ran":     4 },
  4:  { "slowly": 5, "noisily": 5, "EOI": OK },
  5:  { "EOI":    OK }
}

def getNextState(currentState, nextWord):
    return states.get(currentState,{}).get(nextWord, ERROR)

def stateMachine(sentence):
    return reduce(getNextState, sentence, 0) == OK
~~~

PR #36, opened 2020-04-05
Corey Ellis 47612218

Corey23:


README.md
```diff
@@ -42,7 +42,11 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> ASA
+S -> A
+A -> "r"
+A -> "g"
+A -> "b"
```
pragdave, line 49: You're on the right track, but the two A's in the first line don't have to have the same value. (if they did, a grammar rule such as `number = digit+` would only accept numbers where all the digits were the same.

You need something like:

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

# -3

README.md
```diff
@@ -145,7 +166,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+main.cpp
```
pragdave, line 169: Thanks for the set of tests. My only problem with them is that they just output a stream of zeros and ones. They rely on a human to both read them and to match them against the correct results. That's what computers are good at: an automated test should test expected results against actual results, and report when they differ.

# -2

My other problem with the code is that you don't actually have an explicit state machine in there, nor do you have the states represented as a table. You have a parser, but it's really just the same manual code that you'd have written before this class. I was looking for something like this:

~~~ cpp
typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

# -6

README.md
```diff
@@ -164,4 +185,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Type 3 level grammar because each nonterminal goes to a terminal or a terminal followed by a nonterminal which is the formal definition of a type 3 grammar and resembles regular expressions.
```
pragdave, line 188: Nice answer

PR #38, opened 2020-04-05
Assignment 9, Mitchell Morrison

mitchmorrison2:
Mitchell Morrison , #47634734

Please make sure to download the README because my answers from Q1 use angle brackets (<>) which do not show up properly on github UI.


README.md
```diff
@@ -42,7 +42,13 @@ represent the tiles.
...
 ## A1.1
 
-«replace this with your answer»
+S -> r | g | b
+
+S -> r S r
+
+S -> g S g
+
+S -> b S b
```
pragdave, line 51: perfect

README.md
```diff
@@ -53,7 +59,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n) because it should know where it is in the list using 2 pointers (start and end). This will tell you if it meets the requirements of the configuration as it iterates thorugh the list in O(n/2) -> O(n).
```
pragdave, line 62: # -1

Yes, it's O(n). But the parser doesn't know the whole path. The parser takes its input one token at a time. It doesn't know where the end is

README.md
```diff
@@ -109,8 +132,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+StateDiagram.jpeg
```
pragdave, line 135: Nice: you noticed that you can stay in S1 when you see adjectives.

README.md
```diff
@@ -145,16 +178,17 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
-
+Wrote main.cpp up top. 
+run with :
+g++ main.cpp and then ./a.out
```
pragdave, line 183: Thank you for having tests, both positive and negative.

However, they didn't reveal a fairly large problem with the code. A sentence such as "the rabid foaming dog ate angrily" pass the tests. This is because you implemented the parser manually, rather than using a table as the question requested. Your code in (for example) state 1 should return false if neither of the `if` statements triggers, and instead just skips to the next word.

~~~ cpp
state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~

So -6 for not using a table, and -2 for the bugs:

# -8

README.md
```diff
@@ -164,4 +198,5 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+A Finite State Machine is a characteristic of type 3 chomsky grammar.
```
pragdave, line 201: Good answer, and thanks for the cite

PR #40, opened 2020-04-05
Sage

skonstanzer:
Sage Konstanzer
47659629


README.md
```diff
@@ -42,8 +42,16 @@ represent the tiles.
...
+S -> <empty>
+S -> R
+S -> G
+S -> B
+S -> RSR
+S -> GSG
+S -> BSB
+R -> 'r'
+G -> 'g'
+B -> 'b'
```
pragdave, line 54: You're on the right track, but allowing S to be empty breaks it, because the path has to be an odd number of pavers, and these rules allow (for example) "rr"

# -2

also, your can just use r, g, and b as terminals, so

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

README.md
```diff
@@ -53,8 +61,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+A parser for this grammar would likely be roughly O(nlogn), because the runtime grows in proportion to the number of tiles, but not exactly one-to-one - The parser will sometimes need to parse three nonterminals in place of one, such as in the rule S -> RSR.
```
pragdave, line 64: I think your were thinking about runtime, and not memory, although I suspect your estimate is low: I think the runtime will be at least O(n<sup>2</sup>).

# -3

README.md
```diff
@@ -82,17 +89,30 @@ Write the BNF (not EBNF) description for this language.
...
+<sentence> ::= "the" <adjective> <noun> <verb> "."
+<sentence> ::= "the" <adjective> <noun> <verb> <adverb> "."
+<noun> ::= "dog"
+<noun> ::= "cat"
+<verb> ::= "ate"
+<verb> ::= "ran"
+<adverb> ::= "slowly"
+<adverb> ::= "noisily"
+<adjective> ::= "lazy"
+<adjective> ::= "smelly"
```
pragdave, line 103: I think this is mostly correct, but it doesn't allow for repeated adjectives.

# -2

You can also save yourself some typing by using alternatives, and by pushing the decision about optional elements down the grammar:

~~~
  <noun> ::= "dog" | "cat"
  <verb> ::= "ate" | "ran"
~~~

README.md
```diff
@@ -125,8 +144,18 @@ Current state | Next word | Next state
...
+    S0        |    the    |     S1
+    S1        |<adjective>|     S2
+    S1        |  <noun>   |     S3
+    S2        |<adjective>|     S2
+    S2        |  <noun>   |     S3
+    S3        |  <verb>   |     S4
+    S4        |    "."    |     S5
+    S4        | <adverb>  |     S6
+    S6        |    "."    |     S5
+    S5        |    EOI    |    END
```
pragdave, line 158: You have an extra state here: you can combine S1 and S2, so that S1 loops back to itself shen it sees an adjective

# -2

README.md
```diff
@@ -145,7 +174,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+q2-5 the dog ran EOI
```
pragdave, line 177: I don't see any tests here, so 

# -2

And... the question asked for you to encode the table into your programming language. I was expecting to see a data structure that was then traversed by your state machine code (we talked about this in class on Wednesday).

Something like

~~~ cpp
state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

# -6

README.md
```diff
@@ -154,8 +183,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
-
+There are 80 different valid sentences.
```
pragdave, line 186: # -2

Actually, it's infinite, because there can be an indefinite number of adjectives.

README.md
```diff
@@ -164,4 +192,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest Chomsky grammar level for this language is Chomsky Type-2 because in Type-3, the right side of a production is always either a terminal, or a nonterminal followed by a terminal, but in this language the definitions of a sentence contain multiple nonterminals.
```
pragdave, line 195: # -3

But you've already shown it can be implemented using an FSM, so it must be type 3...

pragdave:
In general never add derived files (executables, .obj, etc) to a repository.
They are typically useless to anyone else, because they are build for your
particular environment.

PR #41, opened 2020-04-05
Completed Grammars Homework

AbrahamA00:
Austin Abraham 47661020


README.md
```diff
@@ -42,7 +42,11 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> r | g | b
+
+S -> rSr
+S -> gSg
+S -> bSb
```
pragdave, line 49: Perfect

README.md
```diff
@@ -53,7 +57,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n), because it can load up a stack and push and pop all in the same run, with the value in the center being the only one not checked.
```
pragdave, line 61: Good answer

README.md
```diff
@@ -82,7 +86,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<sentence> ::= "the" <adjective> <noun> <verb> <adverb>
+    <adjective> ::= "smelly" | "lazy" | ""
+    <noun> ::= "dog" | "cat"
+    <verb> ::= "ate" | "ran"
+    <adverb> ::= "slowly" | "noisily" | ""
```
pragdave, line 93: Close: only problem is that it doesn't allow the adjective to repeat

# -2

README.md
```diff
@@ -91,7 +99,11 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+<sentence> ::= "the" {<adjective>} <noun> <verb> [<adverb>]
+    <adjective> ::= "smelly" | "lazy" | ""
+    <noun> ::= "dog" | "cat"
+    <verb> ::= "ate" | "ran"
+    <adverb> ::= "slowly" | "noisily" | ""
```
pragdave, line 106: You don't need the | "" case for adverb, because it is optional

# -1

README.md
```diff
@@ -109,8 +121,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+AbrahamFSM.JPG
```
pragdave, line 124: Only small problem is that an extra state crept in: you can combine S1 and S2

image

# -2

README.md
```diff
@@ -145,7 +166,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+node A25.js
```
pragdave, line 169: First, thank you for giving me both positive and negative test, and for making it clear which are which. The only thing I'd ask of you in future is to make the program itself know which should succeed and fail, so that it can automatically report bugs when an actual result isn't the same as the expected.

Unfortunately, I do have a problem with your FSM. The question asked you to express the table in your language and then write an interpreter that used it. That's ridiculously easy in JavaScript:

~~~ js
states = {
  0: { "the": 1 },
  1: { "lazy": 1, "smelly": 1, "dog": 2, "cat": 2 },
  2: { "ate": 3, "ran": 3 },
  3: { "slowly": 4, "noisily": 4, "EOI": END },
  4: { "EOI": END }
};

function getNextState(currentState, nextWord) {
  return currentState && states[currentState][nextWord]
}

function stateMachine(sentence) {
  return sentence.reduce(getNextState, 0)
}
~~~

Your code doesn't actually use states, except as documentation. 

As a result it doesn't actually parse according to your table: it doesn't allow repeating adjectives

# -8

README.md
```diff
@@ -154,14 +175,19 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+Adjectives: four options (none, lazy, smelly, lazy and smelly)
```
pragdave, line 178: Probably true for your interpreter, but the language itself has an infinite number of sentences

# -3

PR #42, opened 2020-04-05
Homework: Grammar

Kryptsm:
ID: 47604613
Name: Sam Walsh


Kryptsm:
Reupload

README.md
```diff
@@ -18,50 +18,51 @@ Please do not edit anything outside the answers sections.
...
-In one sentence, explain why.
+Q1.1 (5 points)
 
-## A1.2
+Write a Chomsky type 2 grammar that describes any valid path containing at least one paver. Use S as the start state, and r, g, b as terminals that represent the tiles.
 
-«replace this with your answer»
+A1.1
 
+S -> r S r S -> g S g S -> b S b S -> r S -> g S -> b
```
pragdave, line 51: I assume you meant
~~~
S -> r S r 
S -> g S g 
S -> b S b 
S -> r 
S -> g 
S -> b
~~~

which is correct

README.md
```diff
@@ -18,50 +18,51 @@ Please do not edit anything outside the answers sections.
...
-# Q2
+Q1.2 (1 point for the O() answer, 2 for the sentence)
 
-A simple sentence structure might be
+Using big-O notation, what is the likely memory requirement for a parser that can validate a particular path configuration, where n is the number of tiles? In one sentence, explain why.
 
-    "the" (zero or more adjectives) noun verb (optional adverb)
+A1.2
+
+The big-O notation is O(n^3 * |G|) where G is the number of rules. As such big-O of the above configuration is O(6n^3) which simplifies to O(n^3).
```
pragdave, line 59: That's for the runtime, but I asked for memory (which is O(n), as the stack has to hold at least half the tiles).

# -3

README.md
```diff
@@ -72,96 +73,105 @@ Adverbs are "slowly" and "noisily"
...
-«replace this with your answer»
-
+A2.2
 
-## Q2.3 (6 points)
+<<<<<<< HEAD
+sentence ::= "the", {<adjective>}, <noun>, <verb>, [<adverb>];
+=======
+sentence ::= "the", {}, , , [];
+>>>>>>> 07f1444fe496e18ef7278164002acd9890749777
```
pragdave, line 102: Merge failure. A little tacky, young padawan.

This is good as far as it goes, but  you still need to define the non terminals

# -2

README.md
```diff
@@ -72,96 +73,105 @@ Adverbs are "slowly" and "noisily"
...
-  (Hint: my answer has seven states including the start and end states)
+You can hand draw it and snap a picture, or use a drawing tool. Either way, upload the image, and put the file name in the answer below.
 
+(Hint: my answer has seven states including the start and end states)
 
-## A2.3
+A2.3
 
-«replace this with your answer»
+Photo is attached
```
pragdave, line 114: This works, but you have an extra state. You can merge S1 and S2:

image

# -2

README.md
```diff
@@ -72,96 +73,105 @@ Adverbs are "slowly" and "noisily"
...
+>>>>>>> 07f1444fe496e18ef7278164002acd9890749777
+S0        |    the    |   S1
+S1        | Adjective |   S2
+S1        |    Noun   |   S3
+S2        | Adjective |   S2
+S2        |    Noun   |   S3
+S3        |   Verb    |   S4
+S4        |  Adverb   |   S5
+S4        |    EOI    |   END
+S5        |    EOI    |   END
```
pragdave, line 144: This is an accurate representation of your diagram, so no deduction.

README.md
```diff
@@ -72,96 +73,105 @@ Adverbs are "slowly" and "noisily"
...
-The answer should appear in one or more source files in the same directory as
-this file. If I need to do anything more that type a single command to run your
-code, include a script or makefile that will do the job.
+The answer should appear in one or more source files in the same directory as this file. If I need to do anything more that type a single command to run your code, include a script or makefile that will do the job.
 
-## A2.5
+A2.5
 
-«replace this with your answer»
+*See attatched program
```
pragdave, line 157: So, a couple of problems. Although you have three pieces of code that exercise your grammar, they don't qualify as tests. A test knows what the expected result should be, and then complains if the actual result is different. There should be no need for human involvement, except perhaps receiving a notification that something is wrong.

# -2

Technically, you code doesn't parse my grammar, because grammar had smelly and lazy dogs and cat, and you've just implemented it in terms of the nonterminals. If we were writing a general purpose language parser, then this kind of classification-based parse might be useful. I think I'm going to let you get away with it. :)

However, I can't overlook the fact that you didn't actually implement a state machine. What you have here is a simple manual parser.  Here's the kind of state machine I was hoping for:

~~~ cpp
typedef int state;

const int END = 99;

state_table states = {
// current    transitions
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}

    int state = 0;
    for (string word: words) {
        if (state < 0)
            return false;
    }
    return state == 5;
}
~~~

# -6

PR #46, opened 2020-04-05
All parts complete

jdtynes:
Jadon Tynes
47382599


README.md
```diff
@@ -42,7 +42,13 @@ represent the tiles.
...
 ## A1.1
 
-«replace this with your answer»
+S -> r | g | b
+<br/>
+S -> r S r
+<br/>
+S -> g S g
+<br/>
+S -> b S b
```
pragdave, line 51: Perfect (although the <br/>s are a little jarring :)

README.md
```diff
@@ -53,8 +59,11 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+I believe as far as memory goes, it does not matter how many tiles there are,
+<br/>
+it will always be a constant, O(1) because the memory requirement does not grow
+<br/>
+as the number of tiles grow.
```
pragdave, line 66: # -3

So where do you keep the tiles? Remember that the parser sees the input one token at a time, and it will have to store at least half of those tokens to compare against the remainder. So the algorithm is O(n) in memory (and probably O(n<sup<2</sup>) in time.

README.md
```diff
@@ -82,7 +91,15 @@ Write the BNF (not EBNF) description for this language.
...
-«replace this with your answer»
+::= "The" ::= "" | 
+<br/>
+::= "lazy" | "smelly" 
+<br/>
+::= "dog" | "cat" 
+<br/>
+::= "ate" | "ran" 
+<br/>
+::= "" | "slowly." | "noisily."
```
pragdave, line 102: I'm sorry, but this isn't a BNF grammar. And if it were, it doesn't capture the fact that the adjectives can repeat and the adverb is optional.

# -7

README.md
```diff
@@ -91,7 +108,15 @@ Write this grammar using EBNF with common extensions
...
-«replace this with your answer»
+::= "The" * ? 
+<br/>
+::= "lazy" | "smelly" 
+<br/>
+::= "dog" | "cat" 
+<br/>
+::= "ate" | "ran" 
+<br/>
+::= "slowly." | "noisily."
```
pragdave, line 119: # -5

same comment

README.md
```diff
@@ -109,7 +134,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+File: A2-3.pdf
```
pragdave, line 137: The diagram is fine past S2, but the first three states are kind of screwy.

You can't have two transitions from a particular state with the same label (because how could the state machine choose between them) and every transition must be caused by an event.

In your diagram, there are no labels on the transitions from S0, so the state machine has no idea how to get out of it. Then, the adjective transitions have arrows on both ends. 

I was looking for something more like this:

image

# -3

README.md
```diff
@@ -125,7 +150,22 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |     S1
+    S1        |   smelly  |     S2
+    S2        |    dog    |     S3
+    S2        |    cat    |     S3
+    S3        |    ate    |     S4
+    S3        |    ran    |     S4
+    S4        |  slowly   |     S5
+    S4        |  noisily  |     S5
+    S4        |    EOI    |    END
+    S5        |    EOI    |    END
```
pragdave, line 168: This table accurately reflects the diagram, except it gets confused about the double arrows between states 1 and 2. It's incorrect in the big scheme of things, but I'm grading it based on your diagram.

# -2

README.md
```diff
@@ -145,7 +185,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+node A2-5.js
```
pragdave, line 188: First, thank you for writing tests, and including both positive and negative tests.

I added a couple of my own, and confirmed that your parser doesn't support repeating adjectives "the smelly smelly dog"

 I also appreciated that you used a real table to drive it.

However, because your original state machine wasn't really a state machine, you ran into problems, and ended up putting parsing logic in the code, too:

~~~ js
                if (nextState.length == 2
                    &&
                    (words[index+1] == "dog"
                    || words[index+1] == "cat"))
                    currentState = nextState[1]
                else
                    currentState = nextState[0]
                break
~~~

# -4

README.md
```diff
@@ -154,7 +194,11 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+Lazy and/or smelly create 5 possible combinations. Dog or cat is 2 combinations.
+<br/>
+Ate or run is 2 combinations, and slowly or noisily is 3 combinations because it
+<br/>
+could be absent. So, 5*2*2*3 = 60 different valid sentences.
```
pragdave, line 201: # -1

In the original grammar there are an infinite number (because the adjectives can repeat indefinitely. However in your grammar, your answer is true).

README.md
```diff
@@ -164,4 +208,6 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+The simplest would definitely be type 3, because it is easily parseable in a 
+<br/>
+finite state machine.
```
pragdave, line 213: Perfect

pragdave:
Please feel free to reach out if you feel like you've having trouble with an assignment.

PR #48, opened 2020-04-06
Update README.md

jathwe:
Jake Weems 47639050


README.md
```diff
@@ -42,7 +42,16 @@ represent the tiles.
...
+S-> A
+
+S-> A S A
+
+A-> r
+
+A-> g
+
+A-> b
+
```
pragdave, line 54: You're on the right lines, but in "A S A" the two A's do not have to be the same (if they did, then every digit in `number = digit+` would also have to be the same).

You need something like

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

# -3

README.md
```diff
@@ -53,7 +62,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n/2) The stack to parse this must be able to hold half the total number elements as it needs to hold the first half then check it with the second half.
```
pragdave, line 65: Two very minor errors, but no deduction:

1. There's no such thing as O(n/2). You always remove all the low order terms (so O(4n<sup>2</sup>+3n-6) is the same as O(4n<sup>2</sup>), and then you remove constant factors (O(n<sup>2</sup>)

2. In your explanation, you assume the parse knows how long the input is, so that it knows when it reaches the half-way point. It can't.
1 \times 2
$$

README.md
```diff
@@ -82,16 +91,32 @@ Write the BNF (not EBNF) description for this language.
...
+<sentance> ::= "the" [digit{< adjective >}] < noun > < verb > [< adverb >]
+    
+< adjective > ::=  "lazy"|"smelly"
+    
+< noun > ::= "dog"|"cat"
+    
+< verb > ::= "ate"|"ran"
+    
+< adverb > ::= "slowly"|"noisily"
+
```
pragdave, line 103: # -4

This isn't BNF. It's closer to EBNF, but I can't work out what that "digit" means. I was looking for something like

~~~
  <noun> ::= "dog" | "cat"
  <verb> ::= "ate" | "ran"
~~~

README.md
```diff
@@ -109,7 +134,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+fsm.png
```
pragdave, line 137: Your FSM is not valid. 

A state machine has states and transitions that join them. If you are in a state and an event happens, then the state looks for the transition labelled by that event and follows it to the next state.

It looks to me as if you've associated the events with the states themselves. So, for example, when you're in S1, you have two unlabelled transitions to S2 and S3. How is the state machine to know which to take?

I was looking for something like this:

image

# -4

README.md
```diff
@@ -125,7 +150,39 @@ Current state | Next word | Next state
...
+    S1        |    dog    |     S3
+    S1        |    cat    |     S3
+    S2        |    dog    |     S3
+    S2        |    cat    |     S3
+    S3        |    ran    |     S4
+    S3        |    ate    |     S4
+    S4        |    EOI    |     END
+    S4        |   slowly  |     S5
+    S4        |   noisly  |     S5
+    S5        |    EOI    |	END
```
pragdave, line 185: Actually, they're both pretty close. I _did_ want each word spelled out in the table, because that's the way the program is going to want to see it.

The only issue I see with both tables is that you have an unnecessary state: you can merge S1 and S2:

I mentioned in class on Wednesday that the 13 was incorrect, and that my personal answer contained 11 lines.

# -2

README.md
```diff
@@ -145,7 +202,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+I wrote mine in java. As long as JVM is intalled, all that required is that while in this directory, in a terminal type 'java Sentance' and it shall run perfectly with 3 correct and 3 incorrect test cases.
```
pragdave, line 205: And run perfectly it did.

I appreciate you adding passing and failing tests, and that you checked them automatically.

However, your state machine is not table based: you've hard coded it (quite elegantly) as a set of methods.

I was really looking for something that actually had the table in it ("Translate this table into a programming language of your choice"). However, I like your method-based approach, so I'm dinging you less than I would otherwise.

# -3

README.md
```diff
@@ -164,4 +221,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+This would be a level 1 Chomsky. This is because certain non-terminals on either side of terminals are not required but the terminals are required, making this a context-sensitive grammar.
```
pragdave, line 224: But you've already shown you can implement it using a state machine, which makes it type 3...

# -3

PR #52, opened 2020-04-06
Submission

asmith747:
Austin Smith
47029715


README.md
```diff
@@ -42,8 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S ::= r | g | b
+S ::= r S r
+S ::= g S g
+S ::= b S b
```

README.md
```diff
@@ -53,8 +55,9 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+For memory, O(1) is the requirement because we don't need to consider how many
+tiles are in the path or are traversable. We are only checking two tiles at a
+time so we only need to consider two tiles in memory at once.
```
pragdave, line 60: And the other tiles are... where?

The algorithm has to be O(n), because we need to hold the tiles on a stack so we can compare them with those in the rest of the path.

# -3

README.md
```diff
@@ -82,17 +85,20 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
-
+<sentence> ::= "The" <opt-adjective> <noun> <verb> <adverb>
```
pragdave, line 88: So close: this doesn't allow for repeated adjectives

# -2

README.md
```diff
@@ -109,8 +115,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+FSM_Diagram.JPG
```
pragdave, line 118: Beautiful shading in the circles :)

However, it isn't a valid FSM. In an FSM, every states respond to events by following transitions to other states. A state can only have one transition corresponding to a particular event, and every treansition must be caused by an event.

In your diagram, you have "«empty»" between S1 and S2. This can't be right. Think about being at S1.  You have a choice: you can either look at the next token or follow the empty transition. There's no way for the state machine, which has no lookahead, to know what to do. And, if it guesses wrongly, the parse is broken.

You can fix this by simply combining S1 and S2.

# -2

README.md
```diff
@@ -125,8 +130,20 @@ Current state | Next word | Next state
...
+|      S1      |   smelly  |     S1    |
+|      S1      |           |     S2    |
+|      S2      |    cat    |     S3    |
+|      S2      |    dog    |     S3    |
+|      S3      |    ate    |     S4    |
+|      S3      |    ran    |     S4    |
+|      S4      |  slowly   |     S5    |
+|      S4      |  noisily  |     S5    |
+|      S4      |    EOI    |    END    |
+|      S5      |    EOI    |    END    |
```
pragdave, line 146: This is an accurate transcription of your table, so no deduction.

README.md
```diff
@@ -145,17 +162,17 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
-
+create executable:  g++ test.cpp
+run executable:     ./a.out
```
pragdave, line 166: The intent of this exercise was to create an actual data structure containing the table and then use it to drive the interpreter. We discussed this on Wednesday.

You produced a manual parser that really has no connection to the state transitions in the previous two sections.

# -6

test.cpp
```diff
@@ -0,0 +1,78 @@
...
+//<sentence> ::= "The" {"lazy" | "smelly"} ("dog" | "cat") ("ate" | "ran") ["slowly" | "noisily"]
+bool FSM(string sentence) 
+{
+	vector<string> words = tokString(sentence);
+	string adjective[] = { "smelly", "lazy" }, noun[] = { "dog", "cat" }, 
+		verbs[] = { "ate", "ran" }, adverb[] = {"slowly", "noisily"};
+
+	if (words[0] == "the" || words[0] == "The") { //check for initial "the" or "The"
+		int index = 1;
+		while (words[index] != "cat" && words[index] != "dog"){ //keep checking for adjective until word is noun
```
pragdave, line 45: this loops forever if the current word isn't an adjective.

why do you even need the outer loop?

PR #53, opened 2020-04-06
Finished Grammer09

mj3wc:
Wyatt Cronk
ID: 47721548


README.md
```diff
@@ -42,7 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> r|g|b
+S -> r S r
+S -> g S g
+S -> b S b
```

README.md
```diff
@@ -53,7 +56,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(1). This is because you will only need two pointers traversing either end, no extra memory is needed.
```
pragdave, line 59: It you have n tiles, then you need to keep at least half of them in memory, so the algorithm will be O(n).

# -3

README.md
```diff
@@ -82,7 +85,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<adjective> ::= "smelly" | "lazy"
+<noun> ::= "dog" | "cat"
+<verb> ::= "ate" | "ran"
+<adverb> ::= "slowly" | "noisily"
+
+sentence ::= "the" <adjective> <noun> <verb> <adverb>
```
pragdave, line 93: This grammar forces you to have exactly one adjective and one adverb. The original allows zero or more adjectives and zero or one adverb.

~~~
  <noun> ::= "dog" | "cat"
  <verb> ::= "ate" | "ran"
~~~

# -4

README.md
```diff
@@ -91,7 +99,12 @@ Write this grammar using EBNF with common extensions
...
-«replace this with your answer»
+<adjective> ::= "smelly" | "lazy"
+<noun> ::= "dog" | "cat"
+<verb> ::= "ate" | "ran"
+<adverb> ::= "slowly" | "noisily"
+
+sentence ::= "the" {<adjective>} <noun> <verb> [<adverb>]
 
 
 ## Q2.3 (6 points)
```
pragdave, line 110: The diagram FSM.pdf isn't valid: there's a line going from the transition into S1 to state S2.  In a state diagram, each movement between states is triggered by an event, and each state can only have one ongoing transition for a particular event (otherwise, how does it choose which one to take).

The diagram also doesn't allow for multiple adjectives.

You can fix it by removing state S2, and making S1 loop back to itself when it receives an adjective.

# -4

README.md
```diff
@@ -125,7 +138,20 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |    S1
+    S1        |   dog     |    S2
+    S1        |   cat     |    S2
+    S2        |   ate     |    S3
+    S2        |   ran     |    S3
+    S3        |   slowly  |    S4
+    S3        |   noisily |    S4
+    S3        |   EOI     |    END
+    S4        |   EOI     |    END
+
```
pragdave, line 154: This table doesn't agree with your diagram. Interestingly, the table is correct. I'm not sure how you managed that if you interpreted the diagram to produce it...

README.md
```diff
@@ -145,7 +171,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+grammer09.cpp
```
pragdave, line 174: I appreciate you writing tests, but it's really a good idea to have tests where the program reports if they pass or fail. You already have the data there: I just changed line 51 to `if(results[i] != answers[i]){` (and first answers[0], which should be `true`. No deductions for this.

However, the intent of the exercise was to have you use a table to drive the state machine in your code. You wrote a kind of parser by hand, and it does use the states as validation, but as I said on Wednesday, I'm really looking for a table driven approach.

# -6

pragdave:
In general never add derived files (executables, .obj, etc) to a repository.
They are typically useless to anyone else, because they are build for your
particular environment.

PR #54, opened 2020-04-06
Kolodziej Homework Finished

kkolodziej39:


README.md
```diff
@@ -42,7 +42,13 @@ represent the tiles.
...
 ## A1.1
 
-«replace this with your answer»
+S => r
+S => g
+S => b
+S => r S r
+S => b S b
+S => g S g
+
```
pragdave, line 51: Poifect

README.md
```diff
@@ -53,7 +59,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n). One way the parser could work is like checking if a string is a palindrome by setting a pointer at the front and back and incremenet the one at the front and decrement the one at the back until they land on the same character since it needs to have an odd length, checking whether the two are pointing at the same character each time, which would require a checking of n number of tiles. (Also, the parser could first check whether n is even or odd because if it is even, then it would know that it is not valid.)
```
pragdave, line 62: All true, but you're really described a specific algorithm for this particular problem, rather than an actual parse of the grammar itself.

The answer is still O(n), but the parse would likely involve pushing tokens onto a stack, then checking if the remaining tokens are the same as popping those tokens back off. This would be a O(n^2) speed algorithm, but O(n) in memory.

# -1

README.md
```diff
@@ -82,7 +88,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<adjective> ::= "lazy" | "smelly" | ""
+<noun> ::= "dog" | "cat"
+<verb> ::= "ate" | "ran"
+<adverb> ::= "slowly" | "noisily" | ""
+<sentence_structure> ::= "The" <adjective> <noun> <verb> <adverb> "."
```
pragdave, line 95: This is close, but it doesn't allow for multiple adjectives. I was looking for something like

~~~
~~~

# -2

README.md
```diff
@@ -125,7 +139,25 @@ Current state | Next word | Next state
...
+    S1        |  smelly   |     S1
+    S2        |    dog    |     S3
+    S2        |    cat    |     S3
+    S3        |    ate    |     S4
+    S3        |    ate    |     S5
+    S3        |    ran    |     S4
+    S3        |    ran    |     S5
+    S4        |  slowly   |     S5
+    S4        |  noisily  |     S5
+    S5        |    EOI    |     DONE
```
pragdave, line 158: So, imagine you are a state machine interpreter. It's a good steady job; predictable work.

Then someone gives you this table, and asks you to use it to analyze "the dog ran".

Easy, you say. I'm in my start state, S0, and my next word is "the". I just have to find that transition. Wait! There are two. Which do I take? I don't know! I'm deterministic! This isn't fair! 

If this were the original Star Trek, you'd then explode.

The thing I don't understand is that this doesn't correspond at all to your diagram...

# -4

pragdave (commented):
I'm abandonning the grading early after seeing your program

pragdave:
# 0

Your program is a duplicate of another student's.

README.md
```diff
@@ -109,7 +123,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+ProgLangFSM.pdf
```
pragdave, line 126: Good job working out that adjectives keep you in S1

README.md
```diff
@@ -145,16 +177,17 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
 
+Wrote and added a file called "main.cpp"
```
pragdave, line 181: Thanks for writing code to exercise you function. However, they're not really tests. A test is code that knows what it expects, and that report when the actual result is different. There should be no human involvement, except perhaps getting notified that tests have failed.

# -2

And the we have the implementation. The idea of this exercise was to have you write a state machine, ideally a table driven one. That's why it asks you to "Translate this table into a programming language of your choice". 

You've written a manual parser that simply looks at words. There's no mention of states. 

Here's the kind of thing I was looking for:

~~~ cpp
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
// current    transitions
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}

    int state = 0;
    for (string word: words) {
        if (state < 0)
            return false;
    }
    return state == 5;
}
~~~

# -6

pragdave:
All resolved; your grade has not been affected,

PR #55, opened 2020-04-06
done

n-wes:
Nathaniel Wesley 4773313


README.md
```diff
@@ -41,20 +41,21 @@ least one paver. Use `S` as the start state, and `r`, `g`, `b` as terminals that
...
 
 ## A1.1
+```
+S -> r | g | b
 
-«replace this with your answer»
-
-
+S -> r S r | g S g | b S b
+```
```
pragdave, line 48: Nice

README.md
```diff
@@ -41,20 +41,21 @@ least one paver. Use `S` as the start state, and `r`, `g`, `b` as terminals that
...
 can validate a particular path configuration, where `n` is the number of tiles?
 In one sentence, explain why.
 
 ## A1.2
-
-«replace this with your answer»
-
+```
+The parser should require O(n) memory. The recursive case, (S -> r S r | g S g | b S b), will store a tile at each level of the stack. A total of n/2 tiles would be stored in the stack. So O(n/2) -> O(n).
+```
```
pragdave, line 58: # +2 bonus 

for a really nicely thought out answer

README.md
```diff
@@ -82,16 +83,41 @@ Write the BNF (not EBNF) description for this language.
...
+
+                ""
+
+<noun> :: = "dog" | "cat"
+
+<verb> ::= "ate" | "ran"
 
+<adverb> ::= "slowly" | "noisily" | ""
+
+```
```
pragdave, line 101: No deduction: just an observation.

~~~
<adjective> ::= "lazy" | "lazy" <adjective> |
                "smelly" | "smelly" <adjective> |
                ""
~~~

contains some redundancy. It could just be

~~~
<adjective> ::= "" | "lazy" <adjective> | "smelly" <adjective>
~~~

README.md
```diff
@@ -109,8 +135,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+fsm.png
```
pragdave, line 138: I'm giving you the benefit of the doubt and assuming the bottom transition from S4 to S5 was meant to be labelled EOI.

However, I am deducting 2 for the use of "" as a transition between S1 and S2. You can't match nothing in a state machine: that would be ambiguous.

Imagine you're in S1: you could either look at the next token and see if it is an adjective, or you could just shrug and move to S2, breaking the parse.

You can fix this by combining S1 and S2.

# -2

README.md
```diff
@@ -125,7 +150,20 @@ Current state | Next word | Next state
...
+| S1            | "smelly"  | S1         |
+| S1            | ""        | S2         |
+| S2            | "dog"     | S3         |
+| S2            | "cat"     | S3         |
+| S3            | "ate"     | S4         |
+| S3            | "ran"     | S4         |
+| S4            | "slowly"  | S5         |
+| S4            | "noisily" | S5         |
+| S4            | ""        | S5         |
+| S5            | EOI       | END        |
```
pragdave, line 166: # -2

And here, too

README.md
```diff
@@ -145,17 +183,15 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
-
+node sentence.js
```
pragdave, line 186: So... the hack on line 35 where you look ahead should have been a clue that there was something wrong with your table: there's no lookahead in a type 3 grammar.

Also your test `if (curr_state == 6)` should be outside the loop, otherwise you'll return true even if there are words not yet parsed.

Here's what it looks like with the correct state table.

``` js
const END_STATE = Symbol("end state")

const states_tab = {
    0: {
        "The": 1,
    },
    1: {
        "lazy": 1,
        "smelly": 1,
        "dog": 2,
        "cat": 2
    },
    2: {
        "ate": 3,
        "ran": 3
    },
    3: {
        "slowly": 4,
        "noisily": 4,
        "EOI": END_STATE
    },
    4: {
        "EOI": END_STATE
    }
}


let validate_words = (words) => {
    curr_state = 0
    for (let word of words) {
        if (word in states_tab[curr_state]) {
            curr_state = states_tab[curr_state][word]
        }
        else {
            return false
        }

        // END state
        if (curr_state == END_STATE) {
            return true
        }
    }
    return false
}
```

Then you can replace the loop with a call to `reduce`:

~~~ js
const transition = (old_state, next_word) => {
    return old_state && states_tab[old_state][next_word]
}

let validate_words = (words) => {
    return words.reduce(transition, INITIAL_STATE) == END_STATE
}
~~~

# -3 for some silly bugs

PR #56, opened 2020-04-06
Grammars Assignment 

qinyigou:
Zhenbang(John) Zhang
47439058


README.md
```diff
@@ -53,7 +61,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+«big O(n), it is a simply loop to parse every tiles. There is n number of tiles, so it's bigO(n)»
```
pragdave, line 64: # -1

Yes, I think it is O(n), but the loop has nothing to do with memory requirement. It has to push half of the input onto the stack in order to match the other half
in reverse order, so it is O(n)

README.md
```diff
@@ -42,7 +42,15 @@ represent the tiles.
...
-«replace this with your answer»
+«S -> A B C S C B A
+s -> <<empty>>
+ A -> "r"
+ A -> <<empty>>
+ B -> "g"
+ b -> <<empty>>
+ c -> "b"
+ c -> <<empty>>
+ »
```
pragdave, line 53: Good try, but this doesn't work, because there's no guarantee of symmetry. For example, the A, B, and C on the left could expand to a, b, and c, and on the right they could all be empty, giving you the nonsymmetrical result "abc".

You have to approach it from the other end, and say that if S is a valid string, then so is rSr, bSb, and gSg, so the grammar could be

       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b

# -3

README.md
```diff
@@ -82,7 +90,13 @@ Write the BNF (not EBNF) description for this language.
...
 ## A2.1
 
-«replace this with your answer»
+«<sentence>::= <the><adjectives><noun><verb><adverb>
+<the>::= "the"
+<adjectives>::= ""|"lazy"|"smelly"
+<noun>::= "dog"|"cat"
+<verb>::= "ate"|"ran"
+<adverb>::= ""|"slowly"|"noisily"
+»
```
pragdave, line 99: # -2

Close, but your definition of <adjectives> only allows for zero or one of them. It should be something like

~~~
~~~

README.md
```diff
@@ -91,7 +105,9 @@ Write this grammar using EBNF with common extensions
 
 ## A2.2
 
-«replace this with your answer»
+«
+<sentence>::= "the"[""|"lazy"|"smelly"]["dog"|"cat"]["ate"|"ran"][ ""|"slowly"|"noisily"]
```
pragdave, line 109: # -4

This is a good example of where breaking something into multiple lines would help you see the problems

~~~
<sentence>::= "the" [<adjective>] [<noun>] [<verb>] [<adverb>]
<adjective> ::= ""|"lazy"|"smelly"
<noun> ::= "dog"|"cat"
<verb> ::= "ate"|"ran"
<adverb> ::=  ""|"slowly"|"noisily"
~~~

So looking at just the first line, we have "a sentence is the word "the" followed by an optional adjective, an optional noun, an optional verb, and an optional adverb." So "The lazy" is a valid sentence.

~~~
    sentence = "the" adjective* noun verb adverb?
    adjective = "lazy" | "smelly"
    noun = "dog" | "cat"
    verb = "ate" | "ran"
    adverb = "slowly" | "noisily"
~~~

README.md
```diff
@@ -120,13 +136,28 @@ Current state | Next word | Next state
...
+    S1        |   smelly  |     S2
+    S1        |   lazy    |     S2
+    S2        |    dog    |     S3
+    S2        |    cat    |     S3
+    S3        |    ran    |     S4
+    S3        |    ate    |     S4
+    S4        |  noisily  |     S5
+    S4        |           |     S5
+    S4        |   slowly  |     S5
+    S5        |    EOI    |     END
```
pragdave, line 158: You can't have a transition triggered by nothing in a state machine, because that makes it ambiguous: If you're in S1, for example, you can immediately go to S2 without even looking to see if you have the word "the".

This table also doesn't allow for multiple adjectives. And the transition on line 156 should be EOI.

# -3

README.md
```diff
@@ -145,7 +176,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+«TwoPointFive.cpp»
```
pragdave, line 179: 1. The program segv's every now and then (I think when it gets an empty list of words)

2. I'm not sure what the random input helps you test. I ran it maybe 40 times and never saw a valid sentence.

3. I changed the code to test against a valid sentence (` wordList = { "the",  "lazy",  "dog",  "ate",  "slowly" };`) and it failed. It looks like 

     if(wordList[count]== "smelly" && wordList[count] == "lazy"){

can never be true...

# -8

README.md
```diff
@@ -154,7 +185,7 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+«73 valid sentences in TwoPointFive.cpp»
```
pragdave, line 188: In the original grammar, there are infinitely many sentences, as the adjective can repeat forever.

# -3

pragdave:
In general never add derived files (executables, .obj, etc) to a repository. They are typically useless to anyone else, because they are build for your particular environment. In this case there are 47 files checked in, thanks to the IDE...

PR #57, opened 2020-04-06
Finished

j-storey5:


README.md
```diff
@@ -53,8 +55,7 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
-
+Memory required will likely be O(n), because the parser is performing one comparison for each tile.
```
pragdave, line 58: # -1

It is O(n), but because it has to push half of the input onto the stack in order to match the other half
in reverse order, so it is O(n)

README.md
```diff
@@ -42,8 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
-
+S -> r S g S b
```
pragdave, line 45: # -3

Two problems: 

- the grammar will never terminate, because the S terms on the RHS can never resolve to anything other than a nesting of another RHS.

- assuming it could terminate the grammar could produce invalid paths:

~~~ 
S -> r S g S b
       ^ expand
    -> r (r S g S b) g S b
~~~
then assume S is now empty
~~~ 
    ->rrggb
~~~

A possible answer is

~~~
       S -> r | g | b
       S -> r  S  r
       S -> g  S  g
       S -> b  S  b
~~~

See how the first line ensures it can terminate, and the rest ensure that the colors are symmetrical.

README.md
```diff
@@ -82,7 +83,12 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<sentence> ::= "The" adjective noun verb adv
+<adjective> ::= "" | X adj
+<adj> ::= "lazy" | "smelly"
+<noun> ::== "dog" | "cat"
+<adverb> ::= "" | adverb
+<adverb> ::= "slowly" | "noisily"
```
pragdave, line 91: I'm giving you the benefit of the doubt on line 87; I'll assume that "X" was meant to be `<adjective>`.

Lines 90-91 are problematic, though, because they can loop forever without consuming anything.

They need to be written as:

~~~
~~~

# -2

README.md
```diff
@@ -91,7 +97,12 @@ Write this grammar using EBNF with common extensions
...
-«replace this with your answer»
+<sentence> ::= "The" {adjective} noun verb [adverb]
+<adjective>::= "lazy" | "smelly"
+<noun> ::= "dog" | "cat"
+<adverb> ::= "slowly" | "noisily"
+<verb> ::= "ate"|"ran"
+
 
 
 ## Q2.3 (6 points)
```
pragdave, line 108: # <strike>-6</strike>

You gave me a diagram about teleportation...

### Updated when  correct file received

# -3

I understand what you were doing, but this isn't a state transition diagram. They have one circle for each state, and lines from them representing transitions, like this.

image

README.md
```diff
@@ -125,7 +136,23 @@ Current state | Next word | Next state
...
+    S2        |    lazy   |     S2
+    S2        |   smelly  |     S2
+    S2        |    cat    |     S3
+    S2        |    dog    |     S3
+    S3        |    ate    |     S4
+    S3        |    ran    |     S4
+    S4        |   slowly  |     S5
+    S4        |   noisily |     S5
+    S4        |    EOI    |     END
+    S5        |    EOI    |     END
```
pragdave, line 155: This is correct, but contains an unnecessary state: you can remove S2 and simply make S1 loop back to itself when it sees an adjective.
# -2

README.md
```diff
@@ -145,7 +172,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+in javascript file title FSMStorey.js
```
pragdave, line 175: The file I have is called `index.js`.  If that's not the one you intended to send, let me know.

First: I appreciate you submitting a JavaScript program. :)

However, the program doesn't really capture the essence of state transitions. It's a perfectly good manual parser, but it doesn't use the table you'd previously created. I was looking for something like this:

~~~ js
const END = 99;

states = {
  0: { "the": 1 },
  1: { "lazy": 1, "smelly": 1, "dog": 2, "cat": 2 },
  2: { "ate": 3, "ran": 3 },
  3: { "slowly": 4, "noisily": 4, "EOI": END },
  4: { "EOI": END }
};

function getNextState(currentState, nextWord) {
  return states[currentState] && states[currentState][nextWord]
}

function stateMachine(sentence) {
  return sentence.reduce(getNextState, 0)
}
~~~

# -6

README.md
```diff
@@ -164,4 +191,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Chomsky Level 2 because the language requires a list of both terminals and non terminals whose meaning is fixed.
```
pragdave, line 194: # -3

It's actually type 3, because it can be implemented using an FSM (because you can always work out what do do next based on the current state and the next token)

j-storey5:
Professor Thomas,
First of all, thank you for your quick grading and feedback, I messed up when submitting and while I do not expect credit for it, I would like you to know that I did make the finite state machine diagram and it is now uploaded on GitHub, I'll also attach it to this email.  I messed up the file submission and that is 100% my fault and do not expect credit, but if you could still give feedback even if it's ungraded I would appreciate that.
Best,
Jack Storey


________________________________

pragdave:
I still didn't see it on GH, and there was no attachment.

I'll consider grading it if you can get it to me quickly.

j-storey5:
[cid:2c4ea5f5-3027-48f0-a0df-56df75a1d4b2]
________________________________

j-storey5:
Did that attachment finally work?
________________________________

pragdave:
yes. it came through on a different PR

PR #58, opened 2020-04-06
Dillingham Assignment

gracedillingham:
Grace Dillingham, 47639908


README.md
```diff
@@ -42,7 +42,10 @@ represent the tiles.
 
 ## A1.1
 
-«replace this with your answer»
+S -> r | g | b
+S -> r S r
+S -> b S b
+S -> g S g
```
pragdave, line 49: Perfect

README.md
```diff
@@ -53,7 +56,8 @@ In one sentence, explain why.
 
 ## A1.2
 
-«replace this with your answer»
+O(n). This is because it should check starting from both the start and end of the sentence and meet in the middle, 
+which is O(n/2), which is equal to O(n).
```
pragdave, line 61: Yup (although in reality it starts checking from the middle: it builds the stack, checking at each step if the rest of the string matches the stack when popped)

README.md
```diff
@@ -109,7 +121,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
+FSM.png
```
pragdave, line 124: Nice, and in violet, too!

README.md
```diff
@@ -125,8 +137,19 @@ Current state | Next word | Next state
...
+    S1        |   lazy    |     S1
+    S1        |   smelly  |     S1
+    S1        |    dog    |     S2
+    S1        |    cat    |     S2
+    S2        |    ate    |     S3
+    S2        |    ran    |     S3
+    S3        |  slowly   |     S4
+    S3        |  noisily  |     S4
+    S3        |    EOI    |   finish   
+    S4        |    EOI    |   finish
```
pragdave, line 153: Perfect

README.md
```diff
@@ -145,7 +168,7 @@ code, include a script or makefile that will do the job.
 
 ## A2.5
 
-«replace this with your answer»
+main.cpp
```
pragdave, line 171: The objective of this was to express the table you so wonderfully produced in code (as a table). The code you wrote is a fine parser, but it's not using a table: instead it implements the states as explicit code. I was looking for something more like:

~~~ c++
#include <iostream>
using namespace std;
#include <map>

typedef int state;

const int END = 99;

state_table states = {
  { 0, {{ "the", 1 }}},
	{ 5, {{ "EOI", END }}}
};


	if (t.count(nextWord))
		return t[nextWord];
	else
		return -1;
}
~~~

# -3

README.md
```diff
@@ -154,7 +177,8 @@ How many valid sentences are there in this language?
 
 ## A2.6
 
-«replace this with your answer»
+There is an infinite number of sentences possible in this language. This is due to the fact that an infinite number of adjectives
+can be used, and that adjectives can be repeated.
```
pragdave, line 181: So I had to look this up in a grammar book. 

> There is an infinite number of sentences possible in this language.

The subject of the sentence is "sentences", so it should be "There _are_ an infinite number of sentences...".

However, the word "infinite" can also be used as a collective noun, and if used as a noun it is therefore singular, as in "The number of sentences is infinite."

All of which tells you more about my mental state right now than it is about grading....

README.md
```diff
@@ -164,4 +188,4 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+This is a Type 3 because it can be parsed by a finite state machine.
```
pragdave, line 191: Thank you.

pragdave:
In general never add derived files (executables, .obj, etc) to a repository. They are typically useless to anyone else, because they are build for your particular environment.

PR #64, opened 2020-04-06
Finished project for Grammar - 09

kelseyshipman:
Thank you!


README.md
```diff
@@ -82,7 +91,11 @@ Write the BNF (not EBNF) description for this language.
 
 ## A2.1
 
-«replace this with your answer»
+<sentence> ::= <"the"> {<adjective>} {<adjective>} <noun> <verb> {<adverb>} "EOI"
+<adjective> ::= "" | "smelly" | "lazy"
+<noun> ::= "dog" | "cat"
+<verb> ::= "ate" | "ran"
+<adverb> ::= "slowly" | "noisily"
```
pragdave, line 98: # -3

This isn't a pure BNF grammar: BNF doesn't have braces. It also doesn't allow for repetition of adjectives. I was looking for something like

~~~

  <noun> ::= "dog" | "cat"

  <verb> ::= "ate" | "ran"

~~~

README.md
```diff
@@ -109,8 +124,7 @@ Write this grammar using EBNF with common extensions
 
 ## A2.3
 
-«replace this with your answer»
-
+See diagram.jpg
```
pragdave, line 127: # -1

State S2 isn't needed: you can just loop back to S1 if you get an adjective in S1. Otherwise looks great

README.md
```diff
@@ -144,18 +173,18 @@ this file. If I need to do anything more that type a single command to run your
 code, include a script or makefile that will do the job.
 
 ## A2.5
+See grammar_Ans2.cpp
```
pragdave, line 176: This code was failing a few tests. When I looked into it, I found that your state machine code was using incorrect.

The state machine should be taking the words, one at a time, and updating the state. But your original code was using a single variable (`current`) as both the state and the offset to the next word. It was also not checking that it had reached the end state at the end of the sentence, so it would have allowed "the dog ate slowly as the sun set fast in the west."

Here's a corrected version

~~~ c++
bool trySentence(vector<string> words){
  map<int, map<string, int>> states = {
    {0, {{"the", 1}}},
    {1, {{"smelly", 2}, {"lazy", 2}, {"dog", 3}, {"cat", 3}}},
    {2, {{"smelly", 2}, {"lazy", 2}, {"dog", 3}, {"cat", 3}}},
    {3, {{"ate", 4}, {"ran", 4}}},
    {4, {{"slowly", 5}, {"noisily", 5},{"EOI", 99}}},
    {5, {{"EOI", 99}}}
  };
  int state = 0;
  int current_word = 0;
  while (current_word < words.size()){
    state = NextState(state, words[current_word++], states);
    if (state < 0)
      return false;
  }
  return state == 99;
}
~~~

I also can't believe how labor intensive writing those test cases was. I'm giving you some extra points just in sympathy.

# -6

README.md
```diff
@@ -164,4 +193,5 @@ explain why.
 
 ## A2.7
 
-«replace this with your answer»
+Level 3
+The simplest Chomsky grammar level for this language is level 3, as the grammar generates regular language that contains single non-terminal on the left-hand side and a single terminal or single terminal with a single non-terminal on the right-hand side, and the grammar does not perform recursion or backtracking as there are specific states rather than a pattern (like in level 2).
```
pragdave, line 197: Nice answer

PR #65, opened 2020-04-06
Grammar HW

hlambson:


README.md
```diff
@@ -1,167 +1,198 @@
...
+## A1.1
+
+S -> r| g | b
+S -> r S r
+S-> g S g
+S -> b S b
+
+r -> "r"
+g -> "g"
+b -> "b"
```
pragdave, line 52: No deduction, but these aren't technically correct, as r, g, and b are already terminals in the grammar (as they are lower case).

README.md
```diff
@@ -1,167 +1,198 @@
...
+
+  You can hand draw it and snap a picture, or use a drawing tool. Either way,
+  upload the image, and put the file name in the answer below.
+
+  (Hint: my answer has seven states including the start and end states)
+
+
+## A2.3
+
+The file is in github. It is called State Machine.jpg
```
pragdave, line 131: # -2:

The diagram is correct except it doesn't deal with EOI: there's no way of knowing when you've successfully completed. For example, if the sentence was "the dog ate", you'd be in S3. Are we done? How could we know? Just because we ran out of input and we're still in the diagram doesn't mean the sentence is correct (for example the sentence "the dog" also finishes inside the diagram.

You need a DONE state, with EOI transitions from S3 and S4.

README.md
```diff
@@ -1,167 +1,198 @@
...
+    S1        |   smelly  |     S2
+    S1        |   dog     |     S3
+    S1        |   cat     |     S3
+    S2        |   lazy    |     S1
+    S2        |   smelly  |     S1
+    S2        |   ""      |     S1
+    S3        |   ate     |     S4
+    S3        |   ran     |     S4
+    S4        |   slowly  |     S5
+    S4        |   noisily |     S5
```
pragdave, line 159: # -3

This table is not a translation of your diagram. In the diagram. an adjective in S1 keeps you in S1. There's also the problem in the table where you introduce the rule on line 155 to get you back to S1. But you can't match nothing like that: how would the state machine know when to apply the rule.

README.md
```diff
@@ -1,167 +1,198 @@
...
+Then write some unit tests that exercise your function, making sure that it
+recognizes valid sentences and rejects invalid ones.
+
+The answer should appear in one or more source files in the same directory as
+this file. If I need to do anything more that type a single command to run your
+code, include a script or makefile that will do the job.
+
+## A2.5
+
+The files are in the github. It is written in c++, so you just need to run the main.
```
pragdave, line 179: # -12

I tried, honest, but I can't get it to compile. For example, 

~~~ c++
    vector<string> t1 {"the", "cat", "ran", "EOI"};
~~~

just doesn't seem to be valid syntax.

If you can show me this code running, then I'll look again.

hlambson:
Professor Thomas,
I looked at your comments about the homework and for A2.5 you said that you could not get the program to compile. I’ve attached a screenshot of the code running on my laptop and the output that it gives. If you could take another look at it, I would appreciate it.

Thanks,
Henry Lambson

Sent from Mail<https://go.microsoft.com/fwlink/?LinkId=550986> for Windows 10

pragdave:
OK, it took about 30 minutes, but I got it to compile.

You were missing a #include <string> in parser.h, which generated about
1000 lines of random errors. I then had to force my compiler to use C++14
to get the initializers to work.

I will reevaluate it now.

main.cpp
```diff
@@ -0,0 +1,29 @@
...
+
+    parser tester;
+    cout << "A 1 means it is valid, a 0 means it is not" << endl;
+    cout << "Test Sentence 1: " <<  tester.check(t1) << endl;
+    cout << "Test Sentence 2: " <<  tester.check(t2) << endl;
+    cout << "Test Sentence 3: " <<  tester.check(t3) << endl;
+    cout << "Test Sentence 4: " <<  tester.check(t4) << endl;
+    cout << "Test Sentence 5: " <<  tester.check(t5) << endl;
+    cout << "Test Sentence 6: " <<  tester.check(t6) << endl;
+    cout << "Test Sentence 7: " <<  tester.check(t7) << endl;
```
pragdave, line 24: This is not testing. When the program runs, it produces a stream of 1s and 0s, but no indication of what is expected. At the least it should be something like

~~~ c++
typedef vector<string> strings;

void testOneSentence(strings sentence, int expected) {
    parser tester;
    int result = tester.check(sentence);
    cout << ((result == expected) ? "OK" : "ERROR") << ": ";
    for (int i = 0; i < sentence.size(); i++) {
        cout << sentence[i] << " ";
    }
    if (result != expected) {
        cout << " (expected " << expected << ", " << "got " << result << ")";
    }
    cout << endl;
}

int main() {
    vector<strings> testInput = {
      {"the", "dog", "ran", "EOI"},
      {"the", "cat", "ran", "EOI"},
      {"the", "cat", "ran", "slowly", "EOI"},
      {"the", "cat", "EOI"},
      {"the", "smelly", "cat", "ran", "EOI"},
      {"the", "lazy", "lazy", "lazy", "dog", "ran", "noisily", "EOI"},
      {"the", "rabbit", "ran", "really fast", "EOI"}
    };

    int expected[] = {1, 1, 1, 0, 1, 1, 0 };

    for (int i = 0; i < testInput.size(); i++) {
        testOneSentence(testInput[i], expected[i]);
    }
~~~

which gives

~~~
18∙42∙40≻ ./parser
OK: the dog ran EOI 
OK: the cat ran EOI 
OK: the cat ran slowly EOI 
OK: the cat EOI 
OK: the smelly cat ran EOI 
OK: the lazy lazy lazy dog ran noisily EOI 
OK: the rabbit ran really fast EOI 
~~~

Otherwise, you used a map to hold the state table, which is good.

I probably spent an hour in total on this,.

# -4 for the crazy tests and missing #include

